PR #28, opened 2011-02-12
I18n support

etehtsea:
Added
    .human_state
    .human_event_name(event)
methods to support I18n.
Compatible with rails 2 and 3 (tests pass).
Compatible at the moment only with ActiveRecord, but I can rewrite it if another ORM support will be added.

### DIFF
--- a/lib/aasm/i18n.rb
+++ b/lib/aasm/i18n.rb
@@ -3,7 +3,7 @@ class AASM::I18n
     defaults = ancestors_list(klass).map do |ancestor|
       :"#{i18n_scope(klass)}.events.#{i18n_klass(ancestor)}.#{event}"
     end << event.to_s.humanize
-  
+
     I18n.translate(defaults.shift, :default => defaults, :raise => true)
   end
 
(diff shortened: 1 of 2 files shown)


lib/aasm/i18n.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  # added for rails < 3.0.3 compatibility
+  def i18n_klass(klass)
+    klass.model_name.respond_to?(:i18n_key) ? klass.model_name.i18n_key : klass.name.underscore
+  end
+
+  def ancestors_list(klass)
+    klass.ancestors.select do |ancestor|
+      ancestor.respond_to?(:model_name) unless ancestor == ActiveRecord::Base
```
mainameiz, line 33: @etehtsea Hello! Do you remember which problem with `ActiveRecord::Base` was there?

mainameiz: I found that this was introduced in commit https://github.com/mainameiz/aasm/commit/c0d6926a13d9cf4d8e21ca6ec09d6fb99e84538d

Issue #93, opened 2013-10-25
Integer column type support?


bismark64:
I'm wondering if I can do some trick to map an integer column to string in order to use AASM.. 
It'd be wonderful if you add support for integer columns..

PR #124, opened 2014-05-10
Support for edge rails enum fields.

bkon:
This change makes it possible to use integer field declared as Rails 4.1+ enum  to store state.
Seems to be fully backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,8 +309,8 @@ Furthermore, if your column has integer type (which is normally the
 case when you're working with Rails enums), you can omit ```:enum```
 setting --- AASM auto-detects this situation and enabled enum
 support. If anything goes wrong, you can disable enum functionality
-and fall back to the default behavior by setting ```:enum``` to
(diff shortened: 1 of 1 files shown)


mankind:
@alto and @bkon does this feature also allow us to use  **integer column** with **assm** even when we are not using activerecord. I mean can we now do what was described in here: https://github.com/aasm/aasm/issues/93.  

I ask because I am using mongoid and I want to use an integer column with assm instead of string column.

bkon:
@mankind good question.

Unfortunately, it won't work out of the box with anything other than ActiveRecord, as we're using metadata provided by AR to find out what the actual column type is in ```aasm_column_looks_like_enum``` Good news is that's the only AR-specific part of the enum support, so, technically speaking, it could be generalized and pretty easily added to other persistence modules.

mankind:
Thanks for the response @bkon.

PR #145, opened 2014-06-29
added not_ methods to find objects not in the given state

toddp:
aasm has built-in methods to find objects in the specified state; for example, for state 'foo' there is SomeModel.foo  will return a set of records with that state.

I've added similar methods prefixed with not_  , to find objects that are in any state other than the one specified.  For example,  SomeModel.not_foo would return a set of records with state 'bar' or 'woot' but not 'foo'.

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -36,13 +36,9 @@ module AASM
         aasm.current_state == name
       end
 
-
-
       unless @klass.const_defined?("STATE_#{name.to_s.upcase}")
         @klass.const_set("STATE_#{name.to_s.upcase}", name)
       end
-
-
     end
 
     # define an event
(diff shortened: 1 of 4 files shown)


alto:
@toddp How could this issue slip through my focus? I'm sorry – again – for not coming back to this in time.

I think the easiest solution would be to simply define scopes for all kinds, like this:

```ruby
class MyClass < ActiveRecord::Base
  include AASM
  aasm column: 'state' do
    state :requested
    state :confirmed
    state :declined
    ...
  end

  scope :not_declined, -> { 'state <> ?', :declined }
```

and then

```ruby
MyClass.not_declined
```

alto:
Closing this issue for now. Let me know if my suggestions didn't work for you.

PR #160, opened 2014-08-06
Update readme for required to_state argument.

ataber:
The documentation doesn't explain well the fact that when passing arguments to a state transition, the first argument must be the desired end state (or nil). I added some documentation to clarify this.


vergenzt:
> With the upcoming release 4 of _AASM_ this nil parameter will not be required anymore.

Does this mean that line of documentation is no longer accurate for versions 4.x+? :) Could this be reverted if so?

PR #170, opened 2014-09-15
Add :after_commit callbacks to events.

hjdivad:
Event-specific :after_commit callbacks are called in the order that the events
were called.  They are called even when the record is manually persisted.  For
the sake of consistency, the same is now true of the state :after_commit
callback.  As in that case, this feature is specific to active record.

This is really useful when you a) have implicit state transitions; b) have
state-specific validations and  c) have event hooks that you only want to fire
once the transition is committed: for example because they can't be rolled back
(like notifications via a third party).

Imagine if you have something like the following:

``` rb
class Invoice
  aasm do
    state :due
    state :paid, after_commit: :notify_of_payment!

    event :pay do
      transitions from: :due, to: :paid
    end
  end

  with_options if: :paid? do
    validates_presence_of :transaction_id
  end

  def payment_number= n
    write_attribute :payment_number, n
    pay if payment_number_changed?
  end
end
```

You want `notify_of_payment!` to happen on the `pay` transition but only if the
invoice could actually be saved.  When there's state-specific validation it gets
tricky because you want to actually change state _before_ validation so you get
the state-specific validatino, but you want the hook to fire _after_ validation
(indeed after a succesful persistence).

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -56,11 +56,13 @@ module AASM
 
       @klass.send(:define_method, "#{name.to_s}!") do |*args, &block|
         aasm.current_event = "#{name.to_s}!".to_sym
+        aasm.events_fired |= [name.to_sym]
         aasm_fire_event(name, {:persist => true}, *args, &block)
       end
 
       @klass.send(:define_method, "#{name.to_s}") do |*args, &block|
         aasm.current_event = name.to_sym
+        aasm.events_fired |= [name.to_sym]
         aasm_fire_event(name, {:persist => false}, *args, &block)
       end
     end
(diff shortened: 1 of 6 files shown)


deeTEEcee:
Hmm.. still quite confusing with the existing ActiveRecord "after_commit" name...are you guys planning to rename that at all?

alto:
@deeTEEcee What's confusing with `:after_commit`? It should make clear, that this hook is called when the transaction is committed and thereby when the record is persisted (reliably). What would you suggest?

hjdivad:
@alto @deeTEEcee also the name `after_commit` matches the event name in rails exactly and has the same lifecycle semantics (ie after the transaction is committed, so a good time to do things that cannot be rolled back, like sending messages)

PR #201, opened 2015-01-01
Updating copyright date

kerrizor:
..because its 2015!


alto:
@kerrizor Thanks a lot, Kerri!

PR #209, opened 2015-02-04
change method to check attribute_name

facetostool:
Some popular gems like 'globalize' had rewrote method 'attributes': https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/instance_methods.rb 10 line.
So, I had one issue with optimization. Globalize rewrote this method and make request for getting new translated attributes from *_translations table. So, when I tried to optimize MyModel.includes(:translation), I had too many request.
So I think it would be good alternative for this issue.


alto:
Nice improvement. Thanks a lot!

PR #217, opened 2015-03-02
reference state machine object and event in invalid transition error so ...

### LINKED ISSUE #213: AASM::InvalidTransition doesnt reference the object
It would be super handy if this error was initialized with a reference to the object and the requested state.

That way we could still construct the same error response but also do something intelligent without having to parse a string when rescuing from the error/exception.

Happy to make this patch myself but wanted to know if it would be considered

awsmsrc:
...end user can intelligently handle

fixes #213 

Happy to take feedback or make changes, pretty simple change but will make this gem far more useful for my use case!

### DIFF
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -129,7 +129,7 @@ private
     end
 
     if AASM::StateMachine[self.class].config.whiny_transitions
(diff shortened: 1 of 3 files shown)


alto:
@awsmsrc Thanks a lot for this pull request. Easy and straighforward, and so helpful!

alto:
I just released version `4.1.1` including your patch.

PR #218, opened 2015-03-03
Do not update unloaded state for Sequel:

godfat:
The problem is that the record might already have a state, which was
just not loaded due to explicit `select'. In this case, AASM should not
try to write the initial state into database.


alto:
Cheers!

alto:
I just released version `4.1.1` including your patch.

godfat:
Thank you very much for notifying me. Now I can stop pointing to
GitHub from our Gemfile.

PR #219, opened 2015-03-04
Passing event arguments into state callbacks

tobithiel:
I found it useful to also pass the arguments of events into the state callbacks. I was hoping to get it merged back if you see a general use for this.

My scenario is that I have a state which always needs some checks to be done (and takes some optional arguments). I could add the checks to every event into that state. A cleaner way would be to also pass the arguments into these callbacks I think.

### DIFF
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -69,8 +69,10 @@ private
       )
 
       if may_fire_to = event.may_fire?(self, *args)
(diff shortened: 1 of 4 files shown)


alto:
@tobithiel Thanks for your pull request. This change was still on my list (and I somehow always forgot about). It's definitely useful!

alto:
I just released version `4.1.1` including your patch!

PR #225, opened 2015-04-02
Using `aasm_skipping_validations` method

yui-knk:


alto:
I fully agree. Small but beautiful! Cheers!

PR #227, opened 2015-04-16
add ability to merge options for a state instead of erase the previous one

antho1404:
When adding a state two times, the second erase the first one, this can be a bit embarrassing in some case, like a global state machine is defined as a concern and each classe add there own specific states or events and so would like to connect to the callbacks of some existing states.

Now it's so possible to cumulate the states like
```
state :created, after_enter: :do_something, before_enter: :before_enter_callback
state :created, after_enter: :do_something_else
```
this will be the same than 
```
state :created, before_enter: before_enter_callback, after_enter: [:do_something, :do_something_else]
```
instead of actually just 
```
state :created, after_enter: :do_something_else
```

### DIFF
--- a/lib/aasm/core/state.rb
+++ b/lib/aasm/core/state.rb
@@ -28,6 +28,24 @@ module AASM::Core
       name.to_s
     end
 
+    # This merge will append new callbacks from the options to the state
+    # if the state already have a after_enter and the options contains a after_enter callback
+    # the state will have the list of callbacks to execute
+    def merge(options={})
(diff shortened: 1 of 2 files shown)


alto:
@antho1404 I like the idea. And thanks for your work so far. Could you please add some tests for this behaviour?

alto:
@antho1404 Closing this PR now. I didn't merge it, as you didn't reply to my request. Feel free to add tests and re-open it.

PR #228, opened 2015-04-16
reuse respond_to? to be able to check (for rails >= 4.x) if the state attribute exists (even if it's not in the model attributes)

antho1404:
When using a state attribute which is not a column in active record model we can't initialize the state. This case can append when using a kind of alias method instead of use the column config when define the aasm or (in my case) using a multi table inheritance gem like active_record-acts_as where the parent class contains the state attribute. In that case, the model don't have this attribute but respond to the getter and setter method.

Of course this fix only for Rails 4+

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -164,9 +164,8 @@ module AASM
         #   foo.aasm_state # => nil
         #
         def aasm_ensure_initial_state
-          # checking via respond_to? does not work in Rails <= 3
-          # if respond_to?(self.class.aasm.attribute_name) && send(self.class.aasm.attribute_name).blank? # Rails 4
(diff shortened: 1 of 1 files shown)


alto:
@antho1404 Thanks for your pull request. It's nice you are trying to solve issues for _AASM_ with gems like [active_record-acts-as](https://github.com/hzamani/active_record-acts_as) or when the actual attribute is not in the `attributes_names` array.

Did you see that the [Travis CI build is failing](https://travis-ci.org/aasm/aasm/builds/58749500)? Did you run the tests at all?

In order to get this patch considered you should fix the existing testing issues and add tests reflecting your changes and intentions (when creating the pull request). It's common practice to only send pull-requests *with tests* (exceptions are only very minor changes like fixes for typos or comment changes).

alto:
@antho1404 Closing this PR now. I didn't merge it, as you didn't reply to my request. Feel free re-open it.

Issue #158, opened 2014-08-04
multiple state machines in the same model


fabioperrella:
Hi, this is a feature request!

It would be great if could be possible to use more than 1 state machine in the same model. 
We use gem state_machine and we have some models with more than 1 state machines.
When we tried to upgrade rails from 4.0 to 4.1, we were blocked by issue https://github.com/pluginaweek/state_machine/issues/295 so we tried to use gem aasm but it was not possible

Valve:
+1

nofxx:
-1 An object should not have more than one 'state'.
Remove to a class, persist it embedded. 
E.g: The classic state_machine car example: is a mess!
It would look way better if you had a Gearbox, Windshild and whatnot classes instead of only a Car object that handle all those states.

dstosik:
:+1: to @nofxx

Valve:
@nofxx, in my case I have over a hundred states within one object, and we have a second, more high-level coarse state, which acts as a grouping for multiple fine-grained states. 

For example:
Object on a heap:

created (high level grouping state), may actually be one of:
 * initialized
 * allocated
destroyed (another high level grouping state), may be one of:
* inaccessible
* garbage collected 
* finalized

So in this example we have 5 states, but they can be grouped into 2 high-level states, that give you an immediate idea of the state.
Our actual business case is much more complex, this example is just an oversimplified approximation.

Thanks

sepsep:
+1

alto:
Sorry for the long delay with this issue. This is definitely a bigger change to _AASM_.

leofischer:
+1

ShallmentMo:
I'm facing this problem, too. But I wanna ask if I can use like this

```
class Order
  # State Machine #1 
  aasm(column: 'status') do
    ...
   event :cancel, :before => => Proc.new { self.class.aasm.attribute_name "status"} do
       ...
   end
  end

  # State Machine #2 
  aasm(column: 'pay_status') do
    ...
      event :pay, :before => => Proc.new { self.class.aasm.attribute_name "pay_status"} do
        ...
      end
  end
```

I just set class.aasm.attribute_name before every event trigger, is it OK, or it will bring some problem?

ashlinchak:
You can't. AASM use only one state per model. We all are waiting for 4.2.0 where AASM will have multi state support.

ShallmentMo:
I wanna know whether it will cause some side-effect

alto:
[...]

PR #240, opened 2015-06-25
Multiple state machines per class

alto:
work in progress (code changes for #158 )


alto:
Example for multiple state machines per class

```ruby
class ComplexActiveRecordExample < ActiveRecord::Base
  include AASM

  aasm :left, :column => 'left' do
    state :one, :initial => true
    state :two
    state :three

    event :increment do
      transitions :from => :one, :to => :two
      transitions :from => :two, :to => :three
    end
    event :reset do
      transitions :from => :three, :to => :one
    end
  end

  aasm :right, :column => 'right' do
    state :alpha, :initial => true
    state :beta
    state :gamma

    event :level_up do
      transitions :from => :alpha, :to => :beta
      transitions :from => :beta, :to => :gamma
    end
    event :level_down do
      transitions :from => :gamma, :to => :beta
      transitions :from => :beta, :to => :alpha
    end
  end
end
```

Usage
```ruby
example = ComplexActiveRecordExample.new
puts example.aasm(:left).current_state
# => :one
puts example.left
# => 'one'
puts example.aasm(:right).current_state
# => :alpha
puts example.right
# => 'alpha'

example.increment!
# => :two
puts example.left
# => 'two'
puts example.aasm(:right).current_state
# => :alpha
puts example.right
# => 'alpha'
```

alto:
When a state or an event is defined in more than one state machine, a warning is issued:

```ruby
Foo: The event name safe_close is already used!
```

PR #244, opened 2015-08-07
Remove @klass.table_name from scope conditions key

yui-knk:
If we remove @klass.table_name, definded scope works well,
because ActiveRecord dose not need table name to define scope.

And `"#{@klass.table_name}.#{@klass.aasm.attribute_name}"` dose
not work well when we set database name to `table_name`.

### DIFF
--- a/lib/aasm/persistence/base.rb
+++ b/lib/aasm/persistence/base.rb
@@ -82,7 +82,7 @@ module AASM
 
     def create_for_active_record(name)
       conditions = {
-        @klass.table_name => { @klass.aasm(@name).attribute_name => name.to_s }
+        @klass.aasm.attribute_name => name.to_s
       }
       if ActiveRecord::VERSION::MAJOR >= 3
         @klass.class_eval do


yui-knk:
Now there are no tests for witch SQL is created or result of scope, should I add test for these ?

alto:
@yui-knk Thanks for your pull request so far. Your contribution is very appreciated.

But what problem are you trying to solve?

yui-knk:
@alto When I try to use multi database, I face a problem.

For example when I use mysql and two database these database names are

* db1
* db2

I defined `User.table_name = "db2.users"` . This is [supported way in Rails](https://github.com/rails/rails/blob/master/activerecord/test/cases/adapters/mysql2/schema_test.rb#L17).
And User has `state :sleeping`.

By the [issue I reported](https://github.com/rails/rails/issues/21326), 

```ruby
[3] pry(main)> User.where("db.users.id" => 1).to_sql
=> "SELECT `users`.* FROM `users` WHERE `db`.`users.id` = 1"
# [3] is invalid SQL because `db`.`users.id`
```

`User.sleeping` generate invalid SQL :cry:

yui-knk:
sgrif says

> We should definitely continue to make sure that this continues to generate invalid SQL 

so I feel we should remove `@klass.table_name`...

alto:
@yui-knk Sorry for not coming back to you earlier (where has all the time gone?). I now understand your problem. Could you please merge back the current _master_ branch into your branch (and resolving the conflicts) and add one or more tests (as you suggested). That would be really helpful, and I will happily merge your PR afterwards.

alto:
@yui-knk Could you please take a look at #289? Did that (already merged) pull-request do the same thing?

yui-knk:
@alto It looks like #289 does not work with Rails 5 (Rails 5.0.0.beta1) and mysql2 gem.

```
# Gemfile.lock

GIT
  remote: git://github.com/aasm/aasm.git
  revision: 82f592a1edbbca4e43f289314252ee39df1f2a23
  specs:
    aasm (4.8.0)

...
    rails (5.0.0.beta1)
...
```

```ruby
# migration
  def change
    create_table :users, force: :cascade do |t|
      t.string :name
      t.string :aasm_state
    end
...
```

```ruby
class User < ActiveRecord::Base
  include AASM

  self.table_name = "rails_4_1_11_development.users"

  aasm do # default column: aasm_state
    state :running

    event :run do
    end

    event :sleep do
      transitions :from => :running, :to => :sleeping
    end
  end
end
```

```shell
$ bundle exec rails c

> User.sleeping
NoMethodError: undefined method `_reflect_on_association' for nil:NilClass
from rails_5_0_0_beta1/vendor/bundle/ruby/2.2.0/gems/activerecord-5.0.0.beta1/lib/active_record/table_metadata.rb:43:in `associated_table'
```

I thinks it is better to add test cases for mysql2 or pq to aasm gem to support this feature if not regressions will occur :crying_cat_face:

alto:
@yui-knk Could you please have a look at the [failing Travis CI build](https://travis-ci.org/aasm/aasm/builds/108461781)?

alto:
Closing this PR now. 

@yui-knk Feel free to re-open at any time.

PR #247, opened 2015-08-17
:pencil: fix typo

jaredsmithse:


alto:
Uhm, do you think that's a typo?

jaredsmithse:
It seemed like it, so i made a suggestion for the edit.

alto:
Hmm, allright, I just talked to a friend of mine, a native English speaker. And he said, it should be "automatic" instead. Do you agree? (reason is, I don't think it has any magic in it ;) )

jaredsmithse:
Thats fine. Neither the original or my suggestion is a real word I think, but 'automagical' is sometimes used in Ruby (more specifically Rails) to describe some things that are automatically done for you in a way that seems like magic. 

Completely looked over what the normal english word should be :stuck_out_tongue:

alto:
Lovely! Thanks a lot!

PR #251, opened 2015-08-25
Updates 'Callbacks' section example to be correct

naveedkakal:
The current example is trying to define a before block inside of a
transition block, which is not a supported callback.  Moving the before
block in to the event block for an example that is supported.


alto:
Indeed. Thanks a lot!

PR #252, opened 2015-09-01
make activerecord persistence support `attr_accessor` fake column

xieyunzi:
`write_attribute` will check from real db column, but sometimes we need fake column like a middle layer, this make things more flexible.


xieyunzi:
i'll add some test

alto:
@xieyunzi I'm afraid I can't follow. What are you trying to achieve?

PR #256, opened 2015-09-07
Fixed missing name method for state object

Funfun:
Small fix for missing call method for state object.


alto:
It's interesting to see that your change doesn't break any test. And – since with yesterday's release of version `4.3.0` – now there seems to be conflict between your PR and _master_. Bummer!

I would like to merge your change. Would like to resolve that conflict, or shall I?

Funfun:
@alto I will check what i can to with syncing, but this is a hot fix. However i could not find a place where i can wrote down some fail spec first, but i do have real example where the fix helped me. so, Tell me if you need specs.

alto:
Cheers!

anilmaurya:
@alto This change is not required . 

When you compare state with name i.e ` s == name` then State `==` method is called , defined in lib/aasm/core/state.rb

```
    def ==(state)
      if state.is_a? Symbol
        name == state
      else
        name == state.name
      end
    end
```
 
@Funfun If you can show me an example where this fixed helped you then it will be helpful to understand why it didn't work as expected.

alto:
@anilmaurya Indeed. Thanks for pointing that out!

PR #265, opened 2015-10-03
Raise 'column_name undefined error' when column not present

### LINKED ISSUE #152: When used with ActiveRecord, AASM does not check that fields are in the backing db
I recently encountered a bug in some code I wrote where I was attempting to access an AASM state from the backing database associated with an ActiveRecord. I hadn't migrated the database, so AASM couldn't find the column in the database. Instead of returning a helpful error, it said, `NoMethodError: undefined method `type' for nil:NilClass`. I discovered that this happens in active_record_persistence.rb:131 when `self.class.columns_hash[self.class.aasm_column.to_s].type` calls `type` on a `nil` value. It would be helpful if AASM checked that the field was in the database before trying to operate on it.

anilmaurya:
When using enum for state column and enum config not set explicitly,
if column is not present in database then "NoMethodError: undefined
method 'type' for nil:NilClass" was raised for transitions which is
hard to debug.
Fix #152

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -96,7 +96,10 @@ module AASM
         end
 
         def aasm_column_looks_like_enum(name=:default)
-          self.class.columns_hash[self.class.aasm(name).attribute_name.to_s].type == :integer
+          column_name = self.class.aasm(name).attribute_name.to_s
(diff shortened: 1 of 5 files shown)


alto:
@anilmaurya Great work (again :) ). Thanks a lot for your help!

alto:
@anilmaurya Just released version `4.6.0` including your changes.

Issue #267, opened 2015-10-06
Can we add callbacks to parent events in child classes?


reidmorrison:
If we create a base class Job that has a before event callback defined:

```ruby
class Job
  include MongoMapper::Document
  include AASM

  key :state, Symbol, default: :queued

  aasm column: :state do
    state :queued, initial: true
    state :running

    event :start, before: :before_start do
      transitions from: :queued, to: :running
    end
  end

  def before_start
    puts "BEFORE START"
  end
end

j1 = Job.new
j1.start

# => BEFORE START
```

Then we create a child class and want to add more callbacks:

```ruby
class Job2 < Job
  aasm do
    event :start, after: :after_start
  end

  def after_start
    puts "AFTER START"
  end
end

j2 = Job2.new
j2.start

# => AASM::InvalidTransition: AASM::InvalidTransition
```

It appears that `event` completely overrides the previous definition for the event so we need to completely respecify the parent event to extend it.

```ruby
class Job2 < Job
  aasm do
    event :start, before: :before_start, after: :after_start  do
      transitions from: :queued, to: :running
    end
  end

  def after_start
    puts "AFTER START"
  end
end

j2 = Job2.new
j2.start

# => BEFORE START
# => AFTER START
```

I am happy to create a pull request if you have any suggestions on how we should go about allowing child classes to modify parent event definitions?

Maybe add a method that can be called outside the event to add callbacks to it?

```ruby
after_transition on: :start, :do => :after_start
```

Or, maybe dynamically define methods in the aasm block for adding callbacks for each event?

```ruby
class Job2 < Job
  aasm do
    after_start :show_started
  end

  def show_started
    puts "AFTER START"
  end
end
```

anilmaurya:
To add another callback in event, you can do 

```
class Job2 < Job
 
 def initialize
    event = self.class.aasm.events.select{|e| e.name == :start}[0]
    event.options[:after] = [:show_started]
  end

  def show_started
    puts "AFTER START"
  end
end
```

reidmorrison:
Unfortunately changing the options in the child class `Job2` also changes the options in class `Job`. As a result when starting `Job` it now fails trying to call `show_started`

```ruby
j1 = Job.new
j1.start
# => BEFORE START
# => NoMethodError: NoMethodError: undefined method `show_started' for #<Job _id: BSON::ObjectId('5620f107a26ec04db6000003'), _type: "Job", state: :running>:Job
```

[...]

PR #269, opened 2015-10-16
Deep clone state machine during inheritance.

reidmorrison:
Allows child classes to extend the parent state machine without modifying the parents state machine.
Enables https://github.com/aasm/aasm/issues/267


reidmorrison:
@alto any thoughts on this one?

It is a critical capability needed before I can release V2 of [rocketjob](http://rocketjob.io). It allows child jobs to add their own callbacks to aasm state machines defined in parent classes. These callbacks are a major feature of this release. See: https://github.com/rocketjob/rocketjob/blob/v2.0/lib/rocket_job/concerns/event_callbacks.rb

Let me know if you have any suggestions, or a better way to expose callbacks for child classes than the above code.

alto:
@reidmorrison Sorry for coming back to you so late.

No, the PR actually looks pretty good. Only concern I have is that you didn't test that by changing the state machine of the subclass you didn't change the state machine of the superclass, too. Could you add that? I'm a happy PR merger then :+1:

alto:
@reidmorrison Thanks for your recent commit. Could you please add a test covering that fix?

Tybot204:
@alto @reidmorrison I added the tests ensuring the superclass  is not modified in a new PR here: https://github.com/aasm/aasm/pull/429

reidmorrison:
Closing this PR in favor of https://github.com/aasm/aasm/pull/429

Issue #253, opened 2015-09-04
Global callbacks


fabn:
It would be nice (mostly for auditing purpose) to have a set of global callbacks in this gem, something like

```ruby
class Job
  include AASM
  aasm before_state_change: :log_it, before_event: :log_it_too do
    state :sleeping, :initial => true
    state :running
    event :run do
      transitions :from => :sleeping, :to => :running
    end
  end

  def log_it
    logger.info "triggered #{aasm.current_event}"
    logger.info "from #{aasm.from_state} to #{aasm.to_state}"
  end
end
```

What do you think?

jeremywadsack:
+1 I was just looking at how to do this so that we can track even transition timing (and other such metrics).

jeremywadsack:
This looks similar to #221 and #108 which is what I really want. Because I want to decorate an existing AASM model (defined in a shared gem) with callbacks (that are specific to the app) but apply the same code all transitions.

rposborne:
+1

alto:
Okay, I see your point. Will work on that from now on.

alto:
Would this be sufficient?

```ruby
  aasm do
    after_all_transitions :do_logging_or_whatever
  end

  def do_logging_or_whatever
    ...
  end
```

In `do_logging_or_whatever` you can access `aasm.from_state`, `aasm.to_state` and `aasm.current_event`.

jeremywadsack:
Yeah, that works for me.

fabn:
That would be good. Imho also a before callback would be nice to have a trace in case something goes wrong during transition.

anilmaurya:
@alto are you working on this feature ? If not then I can help you on this.

alto:
@anilmaurya Thanks for your offer. I'm sorry, I'm already working on it.

anilmaurya:
@alto No problem . Happy coding :smile:

alto:
Please take a look at #270 and let me know if that works for your problem. Will release soon.

fabn:
Awesome job, thanks.

jeremywadsack:
Yeah, that works for me too. Thanks.

alto:
Closed with #270 .

alto:
I just released version `4.4.0` including the change.

Issue #221, opened 2015-03-12
Logging transitions with DRY code?


awendt:
With a fairly complex state machine, we'd like to log transitions so we know how we ended up in the current user state. For each transition, we'll need:

- `aasm.from_state`
- `aasm.to_state`
- `aasm.current_event`

According to the docs:

> During the transition's `:after` callback (and reliably only then) you can access the originating state (the from-state) and the target state (the to state)

This means we'll go from this:

      transitions from:   :initialized,
                  to:     :mismatched

to this:

      transitions from:   :initialized,
                  to:     :mismatched,
                  after:  Proc.new { |*args|
                    track(
                      'from_state'         => aasm.from_state,
                      'into_state'         => aasm.to_state,
                      'statemachine_event' => aasm.current_event
                    )
                  }

**for each transition** which is undesirable from a maintenance perspective. I'm aware of things like `paper_trail` for auditing AR but that does not fit into our current architecture.

So I'm wondering if there's a way to get this done while staying DRY? (I'm leaning toward `alias_method_chain` to add the callback—but there may be another, AASM-supported way I'm not aware of.)

alto:
@awendt There's already a ticket for something similar (take a look at #108). Would that help you?

What you could do now already is to use methods instead of a `Proc`, like this:

```ruby
aasm do
  transitions from: 	:initialized,
              to:   	:mismatched,
              after:	:tracking
end

def	tracking(*args)
  track(
    'from_state'         => aasm.from_state,
    'into_state'         => aasm.to_state,
    'statemachine_event' => aasm.current_event
  )
end
```

awendt:
@alto Thanks for your quick reply. Anything that involves me declaring "I want this code to run after every transition" just _once_ helps here :)

I found #108 but failed to see a connection. After more careful reading I found [this comment](https://github.com/aasm/aasm/issues/108#issuecomment-69034662):

> FYI the reason I raised the original issue was to implement an observer pattern on transitions. (Of course possible to observe the state variable but not the transitions).

So, yes, that exactly describes the issue I'm having.

I think I will switch to symbols and add `alias_method_chain` for the time being but it would be great to have AASM support that in the long run.

waynn:
@awendt How exactly are you using alias_method_chain with this? I'm running into the same problem, are you including it in the `aasm` declaration?

awendt:
@waynn I settled for a duplicated `after:  :track_event` for each transition. Sorry.

alto:
[...]

PR #270, opened 2015-10-16
Support for global transition callbacks

alto:
Example:

```ruby
class Job
  include AASM

  aasm do
    state :sleeping, :initial => true, :before_enter => :do_something
    state :running

    after_all_transitions :log_status_change
    ...
  end

  def log_status_change
    puts "changing from #{aasm.from_state} to #{aasm.to_state} (event: #{aasm.current_event})"
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 4.4.0 (not yet released)
+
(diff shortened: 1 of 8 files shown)


alto:
This should help with #221 and #253

awendt:
Very nice, thanks for this!

nwp:
This is exactly what I need. When do you plan on releasing 4.4?

alto:
@nwp Released on 2015-10-27 :+1:

nwp:
Thank you so much!

PR #272, opened 2015-10-19
Add a new option to the events method to reject specific events before testing for permitted

dubroe:
We had an issue in our codebase where calling `events(permitted: true)` was causing an infinite loop because one of the events had a guard that called a method that called the method calling the above code.

Our solution was to create this fork and pass in the offending event in the `reject` parameter. I'd assume other people may find this handy as well.

Let me know what you think.

### DIFF
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -49,8 +49,7 @@ module AASM
       state = options[:state] || current_state
       events = @instance.class.aasm(@name).events.select {|e| e.transitions_from_state?(state) }
 
-      options[:reject] = Array(options[:reject])
-      events.reject!{|e| (options[:reject] || []).include?(e.name)}
+      events.reject!{|e| Array(options[:reject]).include?(e.name)}
 
       if options[:permitted]
         # filters the results of events_for_current_state so that only those that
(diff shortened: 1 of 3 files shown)


dubroe:
@alto do you think there's a chance this will get merged in? It would be great to not have to use our forked version from now on.

lib/aasm/instance_base.rb
```diff
@@ -49,6 +49,8 @@ def events(options={})
       state = options[:state] || current_state
       events = @instance.class.aasm(@name).events.select {|e| e.transitions_from_state?(state) }
 
+      events.reject!{|e| Array(options[:reject]).include?(e.name)}
```
alto: Could you please change that to

```ruby
options[:reject] = Array(options[:reject])
events.reject! { |e| options[:reject].include?(e.name) }
```

to make it more readable (more spaces) and better performing (making sure it's an array only once)?

alto:
@dubroe @thisduck I'm happy to accept this PR, as it is a tiny addition in order to make life easier. Thanks a lot!

But please take a look at my comment first. I will merge it in afterwards.

dubroe:
@alto I made the change. Thanks very much!

alto:
Released as version `4.4.1`. Thanks again!

dubroe:
Thank you!

PR #276, opened 2015-10-24
Add test cases which validate results of defined scope

yui-knk:


alto:
@yui-knk Thanks a lot! I'm always happy to improve the testing coverage. :+1:

PR #282, opened 2015-11-04
New Callbacks - Transactions, All Events, Ensure

HoyaBoya:
This PR takes advantage of the global callback foundation introduced in https://github.com/aasm/aasm/pull/270 for ```after_all_transitions```.

In particular, it introduces the following new global events.

* ```before_all_events```, invoked before ```event before```
* ```after_all_events```, invoked after ```event after```
* ```error_on_all_events```, invoked after ```event error```
* ```ensure_on_all_events```, invoked after ```event ensure```
* ```before_all_transactions```, invoked before ```event before_transaction```
* ```before_transaction```, invoked before ```before_aasm_fire_event```
* ```after_transaction```, involved after ```event after_commit``` regardless of success
* ```after_all_transactions```, invoked after ```event after_transaction```

For completeness, a new event level callback was added to compliment ```ensure_on_all_events```.

* ```ensure```, invoked on any ```event``` regardless of completion or error.

This allows AASM to handle like a standard ruby before/rescue/ensure.

The inspiration for this carries from #270, which I believe, is to help DRY up our various callbacks in our state machines.

For example, we wanted to track every time we entered and exited a state (we had to hide it from a AASM workflow GUI). The brute force way was to add ```before``` and ```after``` hooks on every event like so and hope we nailed everything.

### Brute Force

Here's what I think it would look like with current AASM.

```
class Job
  include AASM
  aasm do
    state :running
    state :cleaning

    event :run, before: :transitioning_on, after: :transitioning_off, error: transitioning_off  do
    end

    event :clean, before: :transitioning_on, after: :transitioning_off, error: transitioning_off  do
      transitions :from => :running, :to => :cleaning
    end

    event :sleep, before: :transitioning_on, after: :transitioning_off, error: transitioning_off do
    end

    def transitioning_on
      update_columns(is_transitioning: true)
    end

    def transitioning_off
      update_columns(is_transitioning: false)
    end
  end
end
```

### Monkey Patching

We decided to get clever and instead attack it within  ```aasm_fire_event```.

```
    def aasm_fire_event(event_name, options, *args, &block)
      update_columns(is_transitioning: true)
      super
    ensure
      update_columns(is_transitioning: false)
    end
```
### A Better Way ?

But with the work in #270 and this PR, we can do the following and stop monkey patching into AASM.

```
class Job
  include AASM
  aasm do
    before_all_events :transitioning_on
    ensure_on_all_events :transitioning_off

    event :run do
    end

    event :clean do
      transitions :from => :running, :to => :cleaning
    end

    event :sleep do
    end
  end

  def transitioning_on
    update_columns(is_transitioning: true)
  end

  def transitioning_off
    update_columns(is_transitioning: false)
  end
end
```

If we feel this is the right direction, I'll be happy to add follow up commits with the appropriate documentation changes. At this point, I'm just stumbling behind the hard work done in #270.

UPDATE: I changed the direction of this PR significantly, but I am still stumbling....

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,6 +620,17 @@ Since version *3.0.13* AASM supports ActiveRecord transactions. So whenever a tr
 callback or the state update fails, all changes to any database record are rolled back.
 Mongodb does not support transactions.
 
+There are currently 3 transactional callbacks that can be handled on the event, and 2 transactional callbacks for all events.
+
+```ruby
+  event           before_all_transactions
+  event           before_transaction
+  event           aasm_fire_event (within transaction)
+  event           after_commit (if event successful)
+  event           after_transaction
+  event           after_all_transactions
+```
+
 If you want to make sure a depending action happens only after the transaction is committed,
 use the `after_commit` callback, like this:
 
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Here you can see a list of all possible callbacks, together with their order of 
 
 ```ruby
 begin
+  event           before_all_events
   event           before
   event           guards
   transition      guards
@@ -160,8 +161,13 @@ begin
   old_state       after_exit
   new_state       after_enter
   event           after
+  event           after_all_events
 rescue
   event           error
+  event           error_on_all_events
+ensure
+  event           ensure
+  event           ensure_on_all_events
 end
 ```
 
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -75,6 +75,12 @@ private
     begin
       old_state = aasm(state_machine_name).state_object_for_name(aasm(state_machine_name).current_state)
 
+      event.fire_global_callbacks(
+        :before_all_events,
+        self,
+        *process_args(event, aasm(state_machine_name).current_state, *args)
+      )
+
       # new event before callback
       event.fire_callbacks(
         :before,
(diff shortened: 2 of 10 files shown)


HoyaBoya:
Marking as WIP and going back to formula. I think this is totally wrong and the callback needs to be done after transition ```guard``` is performed.

https://github.com/aasm/aasm/pull/282/files#diff-d3722d359b58a81feb134b78a171731aR34

alto:
Looking forward :+1:

HoyaBoya:
Hi @alto, thanks for taking a look at this PR. 

I've revisited the initial approach and with a green build, I'd love to get your early thoughts on the direction of this PR. I don't think a ```before_all_transitions``` is what we want, but with the global framework introduced, it allows us to consider other types. The PR description is updated to reflect this change.

Super excited to be contributing to AASM and getting into the internals. All feedback welcome.

HoyaBoya:
@alto, I've added several commits to extend the transactional support of AASM for both global and per-event callbacks. From looking at the code a bit more, it appears that my example will be within a transaction and isolated. If I wanted to properly indicate that a AASM model is doing something to other DB processes especially in PG MVCC, I'd need to do:

```
class Job
  include AASM
  aasm do
    before_all_transactions :transitioning_on
    after_all_transactions :transitioning_off

    event :run do
    end

    event :clean do
      transitions :from => :running, :to => :cleaning
    end

    event :sleep do
    end
  end

  def transitioning_on
    update_columns(is_transitioning: true)
  end

  def transitioning_off
    update_columns(is_transitioning: false)
  end
end
```

maokomioko:
+1 would be extremely good to have this merged! :)

JeanMertz:
Looking forward to these callbacks :+1:

alto:
@HoyaBoya Great work, I'm happy to merge this. Thanks a lot!

alto:
Released version `4.7.0` including the addition.

Issue #230, opened 2015-05-21
Do transitions lock models?


rschooley:
I am reading an article from someone at Engine Yard that talks about adding concurrency to aasm.  The article was written 5 years ago; is it still necessary to add this behavior in to aasm?

ref:
http://rhnh.net/2010/07/05/acts-as-state-machine-is-not-concurrent

alto:
No, _AASM_ does not lock the database table row. I think it's overhead and not the responsibility of a state machine. _AASM_ operates on a loaded (fetched) object from the database and writes back its new state. In order to cover that concurrency problem, _AASM_ would need to lock the row, re-fetch the current state from the database, set the new state (or fail if the transition is not allowed anymore, because the tractor has been sold in the meantime), and release the lock.

The issue – as can be seen – is easy to solve in your controller (by setting up a transaction before loading the data from the database).

PR #283, opened 2015-11-06
Introduce Pessimistic Locking For AR Persistence

HoyaBoya:
I understand there has been a previous conversation where we felt that it wasn't appropriate to have pessimistic locking within AASM.

https://github.com/aasm/aasm/issues/230

I offer this PR to revisit the conversation. It adds a ```requires_lock``` configuration to AASM. When set to true or some SQL string i.e. ```FOR UPDATE NOWAIT```, it will be passed through to ActiveRecord's ```lock!``` method. This will allow our AASM models to transition knowing that it owns the row lock and not in contention with another resource that is also accessing the model. It also allows any ```guard``` callbacks we have on the AASM model to operate on the latest data, vs potentially un-committed data if in a PostGres MVCC type of environment.

Our AASM project has fallen into the same pattern as this article by Engine Yard, https://blog.engineyard.com/2010/concurrency-and-the-aasm-gem; we are constantly locking an AASM model in a controller, and potentially in a lower level facade/service layer as we introduce non HTTP touch points to our AASM models (Jobs, Console, other POROs, other AASM models). I'm wondering if this pattern gets ameliorated if we tried to offer locking within AASM.

In summary, it's becoming slightly untenable for us to continue to lock at the controller entry point when thee are so many points of contention.

#### An Example

Using the simple Job example from the README with the lock

```
  include AASM

  aasm requires_lock: 'FOR UPDATE NOWAIT' do
    state :running
    state :cleaning

    event :run do
    end

    event :clean do
      transitions :from => :running, :to => :cleaning
    end

    event :sleep do
    end
  end
end
```

####  Locking With No Conflict

Note the FOR UPDATE NOWAIT query reload of the model.

```
irb(main):002:0> j.run!
   (0.8ms)  BEGIN
  Job Load (0.9ms)  SELECT  "jobs".* FROM "jobs"  WHERE "jobs"."id" = $1 LIMIT 1 FOR UPDATE NOWAIT  [["id", 3]]
  SQL (0.3ms)  UPDATE "jobs" SET "aasm_state" = $1, "updated_at" = $2 WHERE "jobs"."id" = 3  [["aasm_state", "running"], ["updated_at", "2015-11-06 17:57:33.939470"]]
   (6.1ms)  COMMIT
=> true
```

#### With A Conflict

console 1
```
irb(main):001:0> Job.find(3).with_lock { sleep 10000000 } 
```

console 2
```
rb(main):005:0> j.run!
   (0.3ms)  BEGIN
  Job Load (1.8ms)  SELECT  "jobs".* FROM "jobs"  WHERE "jobs"."id" = $1 LIMIT 1 FOR UPDATE NOWAIT  [["id", 3]]
PG::Error: ERROR:  could not obtain lock on row in relation "jobs"
: SELECT  "jobs".* FROM "jobs"  WHERE "jobs"."id" = $1 LIMIT 1 FOR UPDATE NOWAIT
   (0.3ms)  ROLLBACK
ActiveRecord::StatementInvalid: PG::Error: ERROR:  could not obtain lock on row in relation "jobs"
: SELECT  "jobs".* FROM "jobs"  WHERE "jobs"."id" = $1 LIMIT 1 FOR UPDATE NOWAIT
```

Related Documents
http://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -693,6 +693,41 @@ end
 
 which then leads to `transaction(:requires_new => false)`, the Rails default.
 
+### Pessimistic Locking
+
+AASM supports [Active Record pessimistic locking via `with_lock`](http://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html#method-i-with_lock) for database persistence layers.
+
+| Option | Purpose |
+| ------ | ------- |
+| `false` (default) | No lock is obtained | |
+| `true` | Obtain a blocking pessimistic lock e.g. `FOR UPDATE` |
+| String | Obtain a lock based on the SQL string e.g. `FOR UPDATE NOWAIT` |
+
+
+```ruby
+class Job < ActiveRecord::Base
+  include AASM
+
+  aasm :requires_lock => true do
+    ...
+  end
+
+  ...
+end
+```
+
+```ruby
+class Job < ActiveRecord::Base
+  include AASM
+
+  aasm :requires_lock => 'FOR UPDATE NOWAIT' do
+    ...
+  end
+
+  ...
+end
+```
+
 
 ### Column name & migration
 
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -165,7 +165,15 @@ module AASM
           end
 
           begin
-            success = options[:persist] ? self.class.transaction(:requires_new => requires_new?(state_machine_name)) { super } : super
+            success = if options[:persist]
+              self.class.transaction(:requires_new => requires_new?(state_machine_name)) do
+                lock!(requires_lock?(state_machine_name)) if requires_lock?(state_machine_name)
+                super
+              end
+            else
+              super
+            end
+
             if options[:persist] && success
               event.fire_callbacks(:after_commit, self, *args)
               event.fire_global_callbacks(:after_all_commits, self, *args)
@@ -182,6 +190,10 @@ module AASM
 
         def requires_new?(state_machine_name)
           AASM::StateMachine[self.class][state_machine_name].config.requires_new_transaction
+        end
+
+        def requires_lock?(state_machine_name)
+          AASM::StateMachine[self.class][state_machine_name].config.requires_lock
         end
 
         def aasm_validate_states
--- a/spec/database.rb
+++ b/spec/database.rb
@@ -17,24 +17,19 @@ ActiveRecord::Migration.suppress_messages do
     t.string "right"
   end
 
-  ActiveRecord::Migration.create_table "validators", :force => true do |t|
-    t.string "name"
-    t.string "status"
-  end
-  ActiveRecord::Migration.create_table "multiple_validators", :force => true do |t|
-    t.string "name"
-    t.string "status"
+  %w(validators multiple_validators).each do |table_name|
+    ActiveRecord::Migration.create_table table_name, :force => true do |t|
+      t.string "name"
+      t.string "status"
+    end
   end
 
-  ActiveRecord::Migration.create_table "transactors", :force => true do |t|
-    t.string "name"
-    t.string "status"
-    t.integer "worker_id"
-  end
-  ActiveRecord::Migration.create_table "multiple_transactors", :force => true do |t|
-    t.string "name"
-    t.string "status"
-    t.integer "worker_id"
+  %w(transactors no_lock_transactors lock_transactors lock_no_wait_transactors multiple_transactors).each do |table_name|
+    ActiveRecord::Migration.create_table table_name, :force => true do |t|
+      t.string "name"
+      t.string "status"
+      t.integer "worker_id"
+    end
   end
 
   ActiveRecord::Migration.create_table "workers", :force => true do |t|
--- a/spec/unit/persistence/active_record_persistence_spec.rb
+++ b/spec/unit/persistence/active_record_persistence_spec.rb
@@ -425,6 +425,39 @@ describe 'transitions with persistence' do
     expect(persistor).not_to be_sleeping
   end
 
+  describe 'pessimistic locking' do
+    let(:worker) { Worker.create!(:name => 'worker', :status => 'sleeping') }
+
+    subject { transactor.run! }
+
+    context 'no lock' do
+      let(:transactor) { NoLockTransactor.create!(:name => 'no_lock_transactor', :worker => worker) }
+
+      it 'should not invoke lock!' do
+        expect(transactor).to_not receive(:lock!)
+        subject
+      end
+    end
+
+    context 'a default lock' do
+      let(:transactor) { LockTransactor.create!(:name => 'lock_transactor', :worker => worker) }
+
+      it 'should invoke lock! with true' do
+        expect(transactor).to receive(:lock!).with(true).and_call_original
+        subject
+      end
+    end
+
+    context 'a FOR UPDATE NOWAIT lock' do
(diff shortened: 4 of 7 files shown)


lib/aasm/persistence/active_record_persistence.rb
```diff
@@ -145,7 +145,14 @@ def aasm_ensure_initial_state
         end
 
         def aasm_fire_event(state_machine_name, name, options, *args, &block)
-          success = options[:persist] ? self.class.transaction(:requires_new => requires_new?(state_machine_name)) { super } : super
+          success = if options[:persist] 
+            self.class.transaction(:requires_new => requires_new?(state_machine_name)) do
+              lock!(requires_lock?(state_machine_name)) if requires_lock?(state_machine_name)
```
jessmartin: Awesome PR! :smile: :+1: :rainbow: We are facing the same issue on a project right now.

Why `lock!(requires_lock?(state_machine_name))` rather than just `lock!`? 

How could `requires_lock?(state_machine_name)` ever be be passed into `lock!` as `false`?

jessmartin: Ah, I read a little further in the PR and I see. Never mind!

README.md
```diff
@@ -657,6 +657,35 @@ end
 
 which then leads to `transaction(:requires_new => false)`, the Rails default.
 
+### Pessmimistic Locking
```
nn2242: **Pessimistic

HoyaBoya:
I wanted to /ping this PR and gather any thoughts and feedback. I particular, I wanted to provide my thoughts on some of the important points that @alto raised in #230.

>  I think it's overhead and not the responsibility of a state machine.

> The issue – as can be seen – is easy to solve in your controller (by setting up a transaction before loading the data from the database).

On this first point, this is true that it is overhead. But the pessimistic locking mechanics in Active Record tend to overlap with Active Record transactions: the implementation of with_lock imposes it's own transaction.

https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activerecord/lib/active_record/locking/pessimistic.rb#L69-L74

If we had to do ```with_lock``` at the controller level, wouldn't it make the transactions mechanics introduced into AASM redundant? We'd be wrapping mini AASM transactions in a bigger transaction at the controller.

https://github.com/aasm/aasm/blob/master/lib/aasm/persistence/active_record_persistence.rb#L148

My feeling would be that if we felt that pessimistic locking didn't belong in AASM, neither does transaction support. They go hand in hand.

On the second point, having controllers provide the lock, I'd argue that for our use of AASM, we put critical business logic into a state machine. That is why we leverage and are enamored with this project. Part of this critical logic is to know how to ```guard``` appropriately across transitions. As with any system that deals with concurrency, pessimistic locking is a fundamental part of protecting against race conditions and insuring that the guard is acting on the canonical data in the database. As such, for any critical AASM implementation, I would ask that the guard be responsible for the pessimistic lock vs trusting some controller, PORO, Side Kiq job, etc to do it. Having the ability / option, to impose a lock in AASM , would greatly help us enforce this pattern, where the lock is concerned with the AASM model itself (you're on your own if you need other models locked).

jessmartin:
> Part of this critical logic is to know how to guard appropriately across transitions. As with any system that deals with concurrency, pessimistic locking is a fundamental part of protecting against race conditions and insuring that the guard is acting on the canonical data in the database.

:+1: this. @HoyaBoya we're already using your branch in production since our AASM models are utilized concurrently by Sidekiq workers.

alto:
@HoyaBoya Hmm, I'm impressed. Your reasoning is just great, easy to follow and understand. And I fully support your key argument, that pessimistic locking belongs to transactions. 

I will add some comments, and will happily merge your PR once I got your feedback on this.

Oh, and by the way, I learned a new word today: _ameliorated_. Never heard before, and I had to look it up. Thanks for that, too!

README.md
```diff
@@ -657,6 +657,35 @@ end
 
 which then leads to `transaction(:requires_new => false)`, the Rails default.
 
+### Pessimistic Locking
+
+TO DO: Add more information. This is for the sake of PR review...
```
alto: Please provide more information :)

README.md
```diff
@@ -657,6 +657,35 @@ end
...
+
+
+```ruby
+class Job < ActiveRecord::Base
+  include AASM
+
+  aasm :requires_lock => true
```
alto: add `do`

README.md
```diff
@@ -657,6 +657,35 @@ end
...
+
+  ...
+end
+```
+
+```ruby
+class Job < ActiveRecord::Base
+  include AASM
+
+  aasm :requires_lock => 'FOR UPDATE NOWAIT'
```
alto: add `do`

lib/aasm/persistence/active_record_persistence.rb
alto: Yeah, looks weird and confusing. But makes sense once you get it.

spec/database.rb
```diff
@@ -17,24 +17,19 @@
...
-    t.string "name"
-    t.string "status"
-    t.integer "worker_id"
+  %w(transactors no_lock_transactors lock_transactors lock_no_wait_transactors multiple_transactors).each do |table_name|
+    ActiveRecord::Migration.create_table table_name, :force => true do |t|
+      t.string "name"
+      t.string "status"
+      t.integer "worker_id"
+    end
   end
```
alto: Nice cleanups.

HoyaBoya:
Thank you @alto for your feedback. It's with great alacrity that I make this contribution. :smile: .

I've incorporated your suggestions on the feedback section (screen shot below). Let me know if you have any further thoughts or suggestions. I'll rebase into one when we're closer to a finished product.

screen shot 2015-12-11 at 11 11 42 am

HoyaBoya:
@alto, I just want to make sure there are no blockers / feedback waiting on me. Do you require that I rebase master into the PR.

spec/unit/persistence/active_record_persistence_spec.rb
```diff
@@ -399,6 +399,39 @@
...
 
+  describe 'pessimistic locking' do
+    let(:worker) { Worker.create!(:name => 'worker', :status => 'sleeping') }
+
+    subject { transactor.run! }
+
+    context 'no lock' do
+      let(:transactor) { NoLockTransactor.create!(:name => 'no_lock_transactor', :worker => worker) }
+
+      it 'should invoke pessimsitic lock with true' do
```
TheDudeWithTheThing: should this say `should NOT invoke...` ?

HoyaBoya: Ah that is correct @TheDudeWithTheThing. Let me correct the spec description.

HoyaBoya: @TheDudeWithTheThing and @alto, I've corrected the spec descriptions in a  follow up commit. Thanks for the feedback.

alto:
@HoyaBoya I'm asking for a rebase: Please!

HoyaBoya:
@alto I've rebased with the latest upstream master and I've squashed to 1 commit.

alto:
@HoyaBoya Cheers!

alto:

jessmartin:
Awesome @alto! Thanks @HoyaBoya.

Anybody tested this out in production yet? We're using our own fork of aasm and I'd love to switch back to using aasm/aasm.

PR #284, opened 2015-11-18
Clarify docs around the after_commit callback

scottjacobsen:
It appears `after_commit` callbacks only fire when using the bang methods. I understand the non-bang methods do not save the model, but I expected the callbacks to fire when the model is actually saved. 

```
item.approve
item.save!
```

But this does fire the callback:

```
item.approve!
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,7 +615,7 @@ callback or the state update fails, all changes to any database record are rolle
 Mongodb does not support transactions.
 
 If you want to make sure a depending action happens only after the transaction is committed,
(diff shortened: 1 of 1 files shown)


alto:
@scottjacobsen Yeah, that makes things clearer. Maybe `after_commit` is not the best name for it, since you (and others already) assumed that this is a regular _ActiveRecord_ `after_commit` hook (which is not true).

Thanks for extending the README on this. I will think about what to do about the naming.

PR #288, opened 2015-12-02
allow multiple states to be defined on one line

### LINKED ISSUE #146: Allow for multiple state definitions in one line (states :sleeping, :running, :cleaning)
This is a feature request for some syntactic sugar. 

I would like to have an elegant way of adding multiple "vanilla" states in one line. And by "vanilla" I mean any state which isn't initial or has callbacks associated with it.

Example usage of the proposed feature: 

```ruby
aasm do
  state :open, initial: true
  states :processed, :cancelled, :flagged, :succeeded, :failed
  # ...
end
```

If you like the idea and you think it might be of value to more people I would like to prepare a PR.

HParker:
resolves #146

I don't know if anyone still wants it, but it was straight forward enough to add.

I hope you don't mind I coopted the first example in the readme to show the new syntax. I am happy to change that if you prefer it done differently.

*Question:* The issue also mentioned being able to refer to multiple states by `states` rather than `state` in the `aasm` block, however `states` as defined in base is already accessible from that scope as we just instance_eval the block against the base class. I was wondering if you would be open to scoping the methods that are callable from the aasm block to some subset of the methods in base. I am happy to create an issue and take a look.

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -79,8 +79,6 @@ module AASM
     # [1..] state
     def state(*args)
       names, options = interpret_state_args(args)
-      puts names.inspect
-      puts options.inspect
       names.each do |name|
         @state_machine.add_state(name, klass, options)
 
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -63,21 +63,39 @@ module AASM
     end
 
     # define a state
-    def state(name, options={})
-      @state_machine.add_state(name, @klass, options)
-
-      if @klass.instance_methods.include?("#{name}?")
-        warn "#{@klass.name}: The aasm state name #{name} is already used!"
+    # args
+    # [0] state
+    # [1] options (or nil)
+    # or
+    # [0] state
+    # [1..] state
+    def state(*args)
+      if args.last.is_a?(Hash) && args.size == 2
+        names = [args.first]
+        options = args.last
+      elsif
+        names = args
+        options = {}
+      else
+        raise "count not parse states: #{args}"
       end
 
-      @klass.class_eval <<-EORUBY, __FILE__, __LINE__ + 1
+      names.each do |name|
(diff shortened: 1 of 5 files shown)


alto:
@HParker Thanks for the pull request, which looks great so far (and I'm looking forward to merge it). A couple of things though:

1. could you please merge the current _master_ back into your fork?
2. could you please create another ticket for the `states` change you mentioned? (it seems like this is getting a DSL change)

Regarding the README, I'm happy to take your change (and maybe add a comment later).

lib/aasm/base.rb
```diff
@@ -63,21 +63,39 @@ def initial_state(new_initial_state=nil)
...
+    # [0] state
+    # [1] options (or nil)
+    # or
+    # [0] state
+    # [1..] state
+    def state(*args)
+      if args.last.is_a?(Hash) && args.size == 2
+        names = [args.first]
+        options = args.last
+      elsif
```
rousisk: is there a condition missing here?

lib/aasm/base.rb
```diff
@@ -63,21 +63,39 @@ def initial_state(new_initial_state=nil)
...
+        names = args
+        options = {}
+      else
+        raise "count not parse states: #{args}"
       end
```
rousisk: I am wondering whether it would make more sense to have what follows as a separate (private) method which is called directly from this method with the correct arguments. 

e.g. lines 74-75 becoming `new_method_name(args.first, args.last)` etc.

rousisk:
looks good, thanks @HParker - will definitely use that when merged :-)

HParker:
@alto and @rousisk Sorry for the delay!

Should be updated I am really confused how the tests were passing before with that missing case in the `if ... elsif ... else`. Hopefully everything should be working now. Let me know if you have any additional feedback.

alto:

PR #290, opened 2015-12-06
refactoring Base#state_with_scope to reduce complexity of the method

jethroo:
targets an issue shown by https://codeclimate.com/github/aasm/aasm/issues/categories/complexity


alto:
Very nice! :+1:

PR #296, opened 2015-12-18
Make It Easier To Extend AASM With Custom Logic Across Models

HoyaBoya:
This PR tries to make it a bit easier to DRY up logic across AASM models. It accomplishes this by letting developers extend `AASM::Base` and then declare their `aasm` blocks with the extended class.

I believe this will come in handy when we offer more global callbacks from #282. 

Currently, when we want to add functionality to AASM, we are re-opening the `AASM::Base` class to add the functionality we want to share across our many models. Here is an example.

```
module AASM
   class Base
    def use_our_modules!
      @klass.class_eval do
        include OurModule
      end
    end

    def foo!
      # do something foo-ey
    end

    def bar!
      # do something bar-ey
    end
end
```

Taking inspiration from a Rails 5 feature called `ApplicationRecord`, https://github.com/rails/rails/pull/22567, the goal of this PR is to let us represent that functionality in our own `AASM ApplicationRecord` and then have our AASM blocks use this, vs injecting ourselves into `AASM::Base`.

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -1,12 +1,13 @@
 module AASM
   class Base
 
-    attr_reader :state_machine
+    attr_reader :klass,
+      :state_machine
 
     def initialize(klass, name, state_machine, options={}, &block)
       @klass = klass
       @name = name
-      # @state_machine = @klass.aasm(@name).state_machine
+      # @state_machine = klass.aasm(@name).state_machine
       @state_machine = state_machine
       @state_machine.config.column ||= (options[:column] || default_column).to_sym
       # @state_machine.config.column = options[:column].to_sym if options[:column] # master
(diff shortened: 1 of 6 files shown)


lib/aasm/base.rb
```diff
@@ -1,12 +1,13 @@
 module AASM
   class Base
 
-    attr_reader :state_machine
+    attr_reader :klass,
```
HoyaBoya: I've created an attr reader for `@klass` to hide a bit of the internals here.

KevinBongart:
Seems like a good idea, well implemented, tested and even documented! I don't see a reason not to, and would definitely be useful for https://github.com/aasm/aasm/pull/282

alto:
@HoyaBoya Great work (again), just as @KevinBongart mentioned. I like the idea, but wouldn't it be better to use something like `aasm :with => CustomAASMBase do...` instead of `aasm_with CustomAASMBase do...`?

HoyaBoya:
@alto and @KevinBongart, I like that change since it leverages the established configuration DSL. I've gone with `with_klass` as the option, which felt more descriptive of the value that needs to be passed in.

One thing of note is the need re-handle nil here.

https://github.com/HoyaBoya/aasm/commit/33d81dee2b336d549fe437b13e21393f83d4ad21#diff-e38675f41d28dbcc888122e9a41ae58fR43

I think it is due to a chicken and the egg problem where configuration is properly populated with either the configured value or default within the AASM::Base instance, but in this case, I am using it to determine the instance to create and hence I don't have the default set.

Happy to tweak this if anyone has any guidance!

alto:
@HoyaBoya I wouldn't worry about the re-handling issue. It's good enough (your PR is perfect). Please rebase though!

HoyaBoya:
@alto This is rebased with master. Let me know if there are any rebase goofs.

alto:
@HoyaBoya Again, thanks a lot for your work! I really appreciate it.

alto:

PR #298, opened 2015-12-25
Delegate Event After_Commit Callbacks When Using ActiveRecord

rringler:
This PR reimplements @hjdivad's [patch](https://github.com/aasm/aasm/pull/170) delegating event `after_commit` callbacks 
when using ActiveRecord.  This allows the `after_commit` callbacks to be fired 
both when the record is persisted immediately (eg., via `#event!` methods) as  well 
as when the record is persisted manually (eg., via the `#event` method, followed by 
a `#save`.)  It also preserves the order of `after_commit` callbacks in the case of 
multiple state transitions.

This (patched) behavior is  more  consistent with how ActiveRecord handles 
`after_commit` callbacks.

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -107,11 +107,13 @@ module AASM
 
       klass.send :define_method, "#{name}!", ->(*args, &block) do
         aasm(:"#{aasm_name}").current_event = :"#{name}!"
+        aasm(:"#{aasm_name}").events_fired |= [[:"#{name}", *args]]
         aasm_fire_event(:"#{aasm_name}", :"#{name}", {:persist => true}, *args, &block)
       end
 
       klass.send :define_method, "#{name}", ->(*args, &block) do
         aasm(:"#{aasm_name}").current_event = :"#{name}"
+        aasm(:"#{aasm_name}").events_fired |= [[:"#{name}", *args]]
         aasm_fire_event(:"#{aasm_name}", :"#{name}", {:persist => false}, *args, &block)
       end
     end
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -1,11 +1,12 @@
 module AASM
   class InstanceBase
 
-    attr_accessor :from_state, :to_state, :current_event
+    attr_accessor :from_state, :to_state, :current_event, :events_fired
 
     def initialize(instance, name=:default) # instance of the class including AASM, name of the state machine
       @instance = instance
       @name = name
+      @events_fired = []
     end
 
     def current_state
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -36,6 +36,8 @@ module AASM
           aasm_ensure_initial_state
         end
 
+        base.after_commit :aasm_after_commit_hooks
+
         # ensure state is in the list of states
         base.validate :aasm_validate_states
       end
@@ -175,8 +177,9 @@ module AASM
             end
 
             if options[:persist] && success
-              event.fire_callbacks(:after_commit, self, *args)
-              event.fire_global_callbacks(:after_all_commits, self, *args)
+              # Delegating to ActiveRecord
+              # event.fire_callbacks(:after_commit, self, *args)
+              # event.fire_global_callbacks(:after_all_commits, self, *args)
             end
           ensure
             if options[:persist]
@@ -186,6 +189,25 @@ module AASM
           end
 
           success
+        end
+
+        def aasm_after_commit_hooks
+          AASM::StateMachine[self.class].keys.each do |state_machine_name|
+            new_state = aasm(state_machine_name).state_object_for_name(aasm(state_machine_name).current_state)
+            new_state.fire_callbacks(:after_commit, self)
+
+            events_fired = aasm(state_machine_name).events_fired
+
+            until events_fired.empty?
+              event_name, *args = events_fired.shift
+              self.class.aasm(state_machine_name).state_machine.events[event_name].fire_callbacks(:after_commit, self, *args)
+              self.class.aasm(state_machine_name).state_machine.events[event_name].fire_global_callbacks(:after_all_commits, self, *args)
+            end
+          end
+        ensure
+          AASM::StateMachine[self.class].keys.each do |state_machine_name|
+            aasm(state_machine_name).events_fired.clear
+          end
         end
 
         def requires_new?(state_machine_name)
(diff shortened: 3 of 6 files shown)


hjdivad:
Hi again @alto!  Sorry for not following up to your [comments](https://github.com/aasm/aasm/pull/170#issuecomment-57232078) on the earlier PR.

It's great that callbacks are on events and our changes don't depend on manually setting state.  I think the bit where we had a bit of miscommunication on the PR was your third point.

> AASM uses its own ActiveRecord transaction to make sure that only after all callbacks have finished successfully the changes are saved to the database. Therefore, AASM is not using ActiveRecord :after_commit life-cycle hooks.

This is true when using the bang-transition methods `record.event!` but it's not true when using the non-bang methods
```rb
record.event
record.save
```
In those cases aasm is relying on `ActiveRecord` to supply its own transaction.  In those cases aasm is *still* firing `before` and `after` events but, without this PR, it won't fire its `after_commit` events which I contend is something of a footgun.

Consider a common example of sending some communication (SMS, email), that can't be rolled back, after a state transition.  Users shouldn't have to care whether the transition occurs with an implicit save (via a bang-method) or with an explicit save: very likely the business logic will remain the same.

Do you see any reason to prefer the current behaviour?

If not let me know and I'll add something to the changelog and readme on this PR.

Thanks again for your time!

hjdivad:
@alto please let me know if my above explanation is unclear; I think I may have failed to get the point across correctly last time so if there's something that needs clarification I'm happy to try and do so.  Thanks ^_^

alto:
@rringler First of all, thanks a lot for your pull request!

@rringler @hjdivad @chinshr Let's continue #313 using this (older) pull request/issue:

For the records, in #313 I wrote:

> The only explanation is that it has nothing to do with _ActiveRecord_ `after_commit` hooks. It only shares the name, not the behaviour. It's basically a _do this –for the currently triggered event– if and when the change has been committed to the database_. Saying this, it's totally event-based. A pure _ActiveRecord_ `save` (including all manual changes you did) is not going through any other _AASM_ callbacks either. An event trigger without `!` (bang) fires all callbacks that are not related to persisting the state change, as it doesn't persist the state. [@alto, 05/02/2016]

But honestly, all of you make me start thinking about this behaviour. I must admit that I'm not using the `:after_commit` hook myself, so maybe I don't feel the same pain as you do. Although –following what I wrote above (in regards of #313)– I'm still not sure, if that's the right thing to do, as _AASM_'s `after_commit` doesn't share the mechanism provided by _ActiveRecord_. But as already so many people (even before all of you) got into that trap, maybe it's time to meet those requirements/expectations and merge the current pull request. Maybe I pointed the whole idea of this callback into the wrong direction.

What do you think, is this a DSL/behaviour change which should require a new major version?

@rringler Could you please merge the current _master_ into your fork?

hjdivad:
@alto I definitely think the current behaviour should change.  There are two high level issues as I see it.

### Current Issues With Unpatched AASM

1. aasm should not have an event name that exactly matches a rails event name but with different semantics.  This is a recipe for confusion.  I would rather align the semantics (as this PR aims to do) but if we don't end up agreeing on that then at least deprecating this name in favour of something that differs from the ActiveRecord event name would make it less likely that people assume the semantics are the same.

2. The semantics in this PR are desirable and not currently possible in an unpatched AASM.

### Motivation for PR Semantics

It's very useful to have some event that is after commit, as in literally after the transaction is committed: I think we agree on that, but just as a reminder, a great example is scheduling user messaging: you won't want to do this if the transaction is rolled back as you'd wind up SMSing or emailing or whatevering the user with a message that doesn't actually reflect the current state of your data.

Now with unpatched AASM you can do this provided that you only ever use the bang methods `event!` and not `event`.  There are two major problems here.

The first, and less serious, is that it is very error-prone.  It isn't natural at all for people to have to remember that one approach (`event!`) will fire a callback, but another (`event; save`) will not: there is no obviously conceptual reason why this should be so.  This is especially true given the name and the documentation talk about the event being fired when the transaction is committed.  When you only use `event` and manually invoke `save`, there is still a transaction and it is still committed, it's just not one initiated by AASM.

To fully explain AASM's behaviour and differentiate it from the natural expectation would require some very detailed docs explaining a complicated set of exceptions: this is usually a sign that there is something that could be improved with the API.

The second issue is more serious because it's not merely error-prone, there is actually no way to work around it.  The current AASM behaviour works very poorly when you have multiple related state machines (eg on different models) which need to have callbacks fired only after their mutual transaction is successfully committed.

Because of [very confusing ActiveRecord nested transaction behaviour](https://github.com/rails/rails/issues/14698)  an `event!` method inside a callback event of another state machine's `event!` will not properly roll back the outer transaction in the case of certain non-exception raising failures (such as validation failures).  With this PR, you can work around this by ensuring that everything happens in a single transaction: ie within your event handlers you would only invoke `event` state transitions.  With unpatched AASM if you try to do this your `after_commit` event hooks will not be fired.


### Suggested Change

It might be worth seeing if there is anybody who *prefers* the current behaviour?  I've given some reasons to prefer the PR behaviour, but is there any reason to prefer the current behaviour?  Assuming not, what I would recommend is that we add a config flag to opt in to the new behaviour (we can update the PR to support this if you agree) and deprecate the old behaviour.  A future major version would then, as you say, drop the old behaviour.

thoughts?

hjdivad:
@alto this issue is still marked `waiting for response`; is there something I could help clarify or failed to address properly?

hjdivad:
hi @alto; is there anything I could do to help move this PR forward?  Thanks!

alto:
@hjdivad Thanks for summarising the benefits of this PR. And I finally agree to your point of view. Unfortunately, @rringler didn't reply yet to my request to merge _master_ back into this PR (to resolve any conflicts). 

I will try to find time to resolve the conflicts myself, and let this PR be part of _AASM_ version `5.0`.

rringler:
@alto:  I did rebase this PR at one point (though it may have fallen behind again.)  I'm not near my computer at the moment, but I could take a look at it again in a few days.

alto:
@rringler That'd be great. Thanks a lot! And no need to hurry. I really really appreciate your help!

rringler:
@alto: In the process of rebasing this, I've realized that I've lost access to the fork that this PR originates from.  Do I need to open a new PR under a new fork?

alto:
@rringler Hmm, good question. Let me try to merge it manually. If that causes too much hassle I come back to you :)

alto:
@rringler I have to add it to the _aasm5_ branch anyhow.

alto:
@rringler Hmm, no it doesn't seem to be that easy to merge the current _master_ into the PR. Even when resolving the conflict, a couple of tests are failing. Maybe you can give it a try, and create a new PR?

Thanks for your help, I really appreciate it!

rringler:
No worries; I had all the updates made, I just wasn't sure about the mechanics of git/github.  #367 has been created to update this PR.

alto:
Closing this PR in favor of #367, which supersedes this one.

PR #304, opened 2016-01-07
Document return value for auto-saved state transitions

vergenzt:
My team just ran into a bug that took a long time to track down where
an auto-saved state transition method unexpectedly swallowed
exceptions. Some documentation on the behavior would have saved a lot
of searching. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,11 +423,13 @@ job.run   # not saved
 job.run!  # saved
 ```
 
-Saving includes running all validations on the `Job` class. If you want make sure
-the state gets saved without running validations (and thereby maybe persisting an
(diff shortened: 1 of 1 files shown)


alto:
@vergenzt Do you suggest the behavior should be different?

vergenzt:
I'm not sure. Either way seems like a reasonable default--I'm just very new to Rails so it took me a while to realize that that behavior (returning false to indicate errors) is common. Just a documentation change is probably fine.

alto:
@vergenzt Okay, I agree. :+1:

vergenzt:
Thanks. :+1:

Issue #150, opened 2014-07-12
Guarded transitions states should not be listed as permissible states


renatorozas:
I just run into a weird behaviour when I tried to get permissible states. If you have a transition guarded, then  the `to_state` should not be listed as a permissible state.

Example:

``` ruby
class Process
  include AASM
  aasm do
    state :sleeping, :initial => true
    state :running
    state :suspended

    event :start do
      transitions :from => :sleeping, :to => :running, guard: :runnable?
    end

    event :stop do
      transitions :from => :running, :to => :suspended
    end
  end

  def runnable?
    false
  end
end

process = Process.new
process.states(:permissible => true)

# Actual result
 => [:running] # :running should not be listed here.

# Expected result
 => []
```

PR #308, opened 2016-01-13
Fixes incorrect permitted states

eebs:
Permitted states now returns the correct states when an event has more than one transition, and if a transition uses a guard. This is similar to #150.

--

I was trying to get a list of permissible states, but my state machine has an event with multiple transitions, where one transition is guarded. Here's a simplified example that was failing prior to this PR:

```ruby
require 'aasm'

class Foo
  include AASM

  attr_accessor :can_run

  def initialize
    @can_run = false
  end

  aasm do
    state :running
    state :dancing

    event :start do
      transitions :from => :sleeping, :to => :running, guard: :runnable?
      transitions :from => :sleeping, :to => :dancing
    end
  end

  def runnable?
    @can_run
  end
end

foo = Foo.new
foo.can_run = false
foo.aasm.states(:permitted => true).map(&:name)

# Actual
=> [:running]

# Expected
=> [:dancing]
```

Following the behavior mentioned in [Transitions](https://github.com/aasm/aasm#transitions),

> In the event of having multiple transitions for an event, the first transition that successfully completes will stop other transitions in the same event from being processed.

`states(permitted: true)` will only return the first state that would be transitioned to, respecting any guards set on the transitions.

All existing specs pass, and I added a spec for the edge case I found.

Let me know if I missed anything!

### DIFF
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -35,11 +35,23 @@ module AASM
 
     def states(options={})
       if options[:permitted]
-        # ugliness level 1000
-        permitted_event_names = events(:permitted => true).map(&:name)
-        transitions = @instance.class.aasm(@name).state_machine.events.values_at(*permitted_event_names).compact.map {|e| e.transitions_from_state(current_state) }
-        tos = transitions.map {|t| t[0] ? t[0].to : nil}.flatten.compact.map(&:to_sym).uniq
-        @instance.class.aasm(@name).states.select {|s| tos.include?(s.name.to_sym)}
+        permitted_events = events(:permitted => true)
+
+        # An array of arrays. Each inner array represents the transitions that
+        # transition from the current state for an event
+        event_transitions = permitted_events.map {|e| e.transitions_from_state(current_state) }
+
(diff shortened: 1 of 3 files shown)


lib/aasm/instance_base.rb
```diff
@@ -35,11 +35,23 @@ def human_state
...
+        to_state_names = event_transitions.map do |transitions|
+          return nil if transitions.empty?
+
+          # Return the :to state of the first transition that is allowed or nil
+          transition = transitions.find { |t| t.allowed?(@instance) }
+          transition ? transition.to : nil
+        end.flatten.compact.uniq
+
+        # Select states that are in to_state_names
+        @instance.class.aasm(@name).states.select {|s| to_state_names.include?(s.name)}
```
alto: Nice. The code looks much better!

alto:
@eebs Great work! Thanks a lot, I'm happy to merge.

eebs:
Thanks! Happy to help!

alto:
Will release soon.

alto:

eebs:
Fantastic, thanks again!

PR #311, opened 2016-01-20
Fix #297 duplicate after_all_transitions after reload

### LINKED ISSUE #297: duplicate global callbacks caused by `after_all_transitions`
#### Base code
Using aasm4.5.0, rails4.2.0

    class Job
      include AASM

      aasm do
        state :sleeping, :initial => true, :before_enter => :do_something
        state :running

        after_all_transitions :log_status_change

        event :run, :after => :notify_somebody do
          before do
            log('Preparing to run')
          end

          transitions :from => :sleeping, :to => :running, :after => Proc.new {|*args| set_process(*args) }
        end
      end

      def log_status_change
        puts "changing from #{aasm.from_state} to #{aasm.to_state} (event: #{aasm.current_event})"
      end
    end

#### Problem
I have added a puts at lib/aasm/base.rb#112

    def after_all_transitions(*callbacks, &block)
      puts '---> add after all transaction'
      @state_machine.add_global_callbacks(:after_all_transitions, *callbacks, &block)
    end

The '---> add after all transaction' will printed when I change some code in the
Job model and rails auto reload Job model.
And when I call `job.run!`. I will get two line of log puts

    changing from ...
    changing from ...

#### Analysis
I checked the `add_global_callbacks` method.
`@global_callbacks[:after_all_transitions]` get values as: `[:log_status_change, :log_status_change]`

    def add_global_callbacks(name, *callbacks, &block)
      @global_callbacks[name] ||= []
      callbacks.each do |callback|
        @global_callbacks[name] << callback
      end
      @global_callbacks[name] << block if block
    end

So I think it is a bug of `after_all_transitions`.

@alto

lingceng:


alto:
@lingceng What's the actual problem you are trying to solve? What kind of _reload_ are you simulating by removing a constant with `Callbacks.send(:remove_const, :Basic)` (as done in the spec you provided)?

lingceng:
@alto  See #297

alto:
@lingceng Ah, I see. Sorry, my fault!

alto:
@lingceng Nice spotting! Thanks for the PR. Happy to merge!

alto:

lingceng:
:+1:

lingceng:
@alto  You'd better set the milestone for this PR. So others can get the info from the milestone tag.

PR #315, opened 2016-01-31
RubyMotion support.

Infotaku:
Hi,

I wanted to be able to use AASM in my new RubyMotion project.
I fooled around a little to see if it was possible and by the time I was done with it I realised all specs were still valid. So I was wondering if you would be interested to merge that support back on your side ?

Basically all I had to do is to remove all string-based eval-like code and replace them with a good old define_method. That and 2 minor changes on the persistence 'constantize' method and the list of required files by the gem. That's all.

So there it is !

### DIFF
--- a/README.md
+++ b/README.md
@@ -750,6 +750,11 @@ Job.aasm.states_for_select
 ```
 
 
+### RubyMotion support
+
+To use AASM with a RubyMotion project, use it with the [motion-bundler](https://github.com/archan937/motion-bundler) gem.
+
+
 ### Testing
 
 AASM provides some matchers for [RSpec](http://rspec.info): `transition_from`, `have_state`, `allow_event` and `allow_transition_to`. Add `require 'aasm/rspec'` to your `spec_helper.rb` file and use them like this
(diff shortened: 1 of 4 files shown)


alto:
@Infotaku Wow, great pull request. Very happy to merge!

alto:
@Infotaku I just wonder how we can make sure that future releases don't break this compatibility. Any ideas?

Infotaku:
@alto Glad you like it. 

RubyMotion does not support string-based runtime eval because it requires a full-blown interpreter. So you should avoid using class_eval, instance_eval and eval with anything other than blocks. That's why I had to change the way AASM defines its state-related methods and use define_method instead.

Also since RubyMotion has to compile the files, runtime require is not supported (you can still use it, but it will have no effect and the code won't be loaded). This is why I had to require the default persistence file (plain-persistence.rb) inside aasm.rb. So the 'motion-bundler' gem knows it has to add the file for compilation. This is not mandatory (motion-bundler offers some logic to specify missing files manually) but it's a nice to have and provides peace of minds for RubyMotion developers.

alto:
@Infotaku Thanks for pointing out. But what I actually meant was to find a way to automatically check/test if support for _RubyMotion_ has been broken with any future code change. Something a Travis CI build can tell us about.

Infotaku:
Well, I guess we would need to setup a RubyMotion project and find a way to run all the (non database related) aasm specs. I do have access to rake with RubyMotion but I don't know how tu run à remote gem specs from a local project.
Also, that would require an OSX worker server for Travis (XCode si required to build the project, which is required to run the specs).

alto:

Infotaku:
By advised that I found an issue on this PR, https://github.com/aasm/aasm/pull/318 fixes it.

PR #316, opened 2016-02-02
Add option to use Class that responds to call for after transitions

mlr:
Hi,

After transitions can sometimes become large or complex enough to warrant their own classes to provide a separation of concerns or to simply organize code in a cleaner way.  A common pattern in ruby/rails projects is to delegate this work to a class that responds to a `call` method.

This PR will allow AASM defined state machines to use the `after` transition option by simply passing the name of a class that responds to `call`. It also allows you to pass in arguments to the after transition, similar to the other `after` options. In this case, the arguments will automatically be provided to the `initialize` method of the class you define.

Example:

```
  include AASM

  aasm do
    state :ready
    state :completed

    event :complete do
      transitions from: :ready, to: :completed, after: CompleteEvent
    end
  end
end

class CompleteEvent
  def initialize(record, notes)
    @record = record  # provides access to the underlying Job instance
    @notes = notes    # contains "Notes about the work that was completed" in this example
  end

  def call
    # Notify the customer that the job is completed,
    # record the total time it took to complete the job
    # or any other work that needs to happen.
  end
end

job.complete!("Notes about the work that was completed")
```

Please let me know your thoughts.

Thanks for maintaining such an excellent project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ class Job
       end
 
       transitions :from => :sleeping, :to => :running, :after => Proc.new {|*args| set_process(*args) }
+      transitions :from => :running, :to => :finished, :after => LogRunTime
     end
 
     event :sleep do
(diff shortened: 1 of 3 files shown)


alto:
@mlr What a great idea!

alto:
Do you mind adding a note to the `README.md`? Happy to merge afterwards!

mlr:
Added a brief example of the option to the README. Let me know if I should add something more thorough that explains the argument usage for classes other than the in-line comment:

```
# optional args parameter can be omitted, but if you define initialize
# you must accept the model instance as the first parameter to it.
```

I began to feel like the examples explaining argument usage in event/transition options could be broken into its own sub section at this point, but wasn't sure I should take liberty to do all that in this PR.

Let me know also if you'd prefer the commits squashed -- thanks!

alto:
I think –besides the typo– this PR is good to go.

alto:
@mlr Forget it. Sorry for making things more complicated than needed. I'm happy to merge this PR NOW!

mlr:
Gah! unfortunate typo, apologies. Thanks anyhow for the merge!

alto:

PR #325, opened 2016-02-18
Fix #273 work right after dup

### LINKED ISSUE #273: Transitions fail on dup'd copies
If you dup a model, the copy won't successfully transition.  Example trivial migration task, model, and rails console lines provided.  Notice that when the copy transition is attempted, it's modifying the status of **the source id**, not the copy id.  It should update id 6, but updates id 5.  The SQL argument is cropped by the size of the available text window.

```rails generate migration CreateAasmBug status:string```
```
    class AasmBug < ActiveRecord::Base
      include AASM

      aasm :column => 'status' do
        state :begin, initial: true
        state :end

        event :finish do transitions from: [:begin], to: :end do end end
        event :restart do transitions from: [:end], to: :begin do end end
      end
    end
```

```
    Loading development environment (Rails 4.2.0)
    [1] pry(main)> source = AasmBug.create!
    => #<AasmBug:0x007ff6ff2c3c08 id: 5, status: "begin">
    [2] pry(main)> source.finish!
    [3] pry(main)> copy = source.dup
    [4] pry(main)> copy.save!
    [5] pry(main)> copy
    => #<AasmBug:0x007ff6fdeb8b78 id: 6, status: "end">
    [6] pry(main)> copy.restart!
       (0.1ms)  BEGIN
      SQL (0.3ms)  UPDATE "aasm_bugs" SET "status" = $1 WHERE "aasm_bugs"."id" = $2  [["status", "begin"], ["id", 5]]
       (1.8ms)  COMMIT
    => true
    [7] pry(main)> copy
    => #<AasmBug:0x007ff6fdeb8b78 id: 6, status: "end">
```

lingceng:
See:
http://www.jonathanleighton.com/articles/2011/initialize_clone-initialize_dup-and-initialize_copy-in-ruby/

### DIFF
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -67,6 +67,11 @@ module AASM
     @aasm[name.to_sym] ||= AASM::InstanceBase.new(self, name.to_sym)
   end
 
+  def initialize_dup(other)
+    @aasm = {}
+    super
+  end
+
 private
 
   # Takes args and a from state and removes the first
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,6 +3,7 @@ $LOAD_PATH.unshift(File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib
 require 'aasm'
 require 'rspec'
 require 'aasm/rspec'
+require 'pry'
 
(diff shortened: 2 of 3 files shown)


alto:
@lingceng Very nice! Thanks a lot!

alto:

PR #326, opened 2016-02-18
Use a symbol as the name of AR callback instead of block

sineed:


sineed:
Hi.

Thanks for `aasm` gem. I use it in my project and find it very useful.
I have some cases when I have an AR object that is not prepared for state machine usage. So I don't want to use the callback defined an AR persistence file and want to skip it:

```ruby
class MyClass
  include AASM
  
  skip_callback :initialize, :after, :aasm_ensure_initial_state
end
```

But now I can't do like that because I need to find an object id of proc object which I find really hard.

This PR solves my problem.

alto:
@sineed Sounds reasonable. Thanks!

alto:

csmuc:
In my case (retrieving many records in an API) the method-based approach is also significantly faster than the old proc version. `Benchmark.ips` shows 22.6 requests/s vs. 20.8 requests/s.

Actually I would like to skip this callback altogether when simply loading stuff from the DB for rendering (23.4 requests/sec when I remove the callback).

PR #338, opened 2016-03-08
update a missed state in README

satour:


alto:
@satour Indeed! Thanks a lot! :)

PR #340, opened 2016-03-15
Fix warning when overriding an exsiting method

### LINKED ISSUE #335: state :valid breaks validations
aasm do
      state :valid
    end

gives me

`wrong number of arguments (1 for 0)`
`My_model#perform_validations`
`activerecord (4.2.5.2) lib/active_record/validations.rb, line 83`

    def perform_validations(options={}) # :nodoc:
      options[:validate] == false || valid?(options[:context])
    end

Using Rails 4.2.4.2 and Ruby 2.2.4

Maybe stupid of me to try to name a state `valid`, but I needed some time to figure it out :( Should be a warning somewhere. Thanks

pirj:
When overriding an existing method in base class (be it event redefinition by another state machine for same class or method originally defined in class itself), there should be a warning message.

The meat: klass.instance_methods.include?(method_name*.to_sym*)
Test added. Some rearrangement.

Fixes #335

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ simple.aasm(:work).current
 
 _AASM_ doesn't prohibit to define the same event in more than one state machine. The
 latest definition "wins" and overrides previous definitions. Nonetheless, a warning is issued:
-`SimpleMultipleExample: The aasm event name run is already used!`.
+`SimpleMultipleExample: overriding method 'run'!`.
 
 All _AASM_ class- and instance-level `aasm` methods accept a state machine selector.
 So, for example, to use inspection on a class level, you have to use
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -74,13 +74,10 @@ module AASM
     def state(name, options={})
       @state_machine.add_state(name, klass, options)
 
-      if klass.instance_methods.include?("#{name}?")
-        warn "#{klass.name}: The aasm state name #{name} is already used!"
-      end
-
-      aasm_name = @name
-      klass.send :define_method, "#{name}?", ->() do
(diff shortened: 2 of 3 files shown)


alto:
@pirj Actually, that's pretty cool. Introducing something like `safely_define_method` is even better. Great PR, thanks a lot!

alto:

reicheltd:
+1 thx

leonelgalan:
Thanks for adding this warning, I came across the this when upgrading aasm. In my case using an an enum to enumerate states before defining the state machine triggers the warning:

```
ModelName: overriding method 'a?'!
ModelName: overriding method 'b?'!
```

```ruby
enum a_status: {
  a: 0,
  b: 1
}

aasm column: a_status do
  state :a
  state :b 
}
```

As far as I can tell, there is no harm done, I just need to figure out what behavior I'm using from enum other than `a?` and `b?` to see if it's worth keeping or not.

pirj:
@leonelgalan Nice catch!

leonelgalan:
Sure, let me study this a bit more. My first impression, is that the new warning was working exactly as expected.  I'm not sure why I created an enum before the state machine, once I figure out I will report if I believe this is still an issue that should be fixed.

For now, I just wanted to annotate my findings, in case anyone else see's the new warning after upgrading.

pirj:
@leonelgalan You did that to store the state as an integer as opposed to storing as a string that is a default, this valid, makes sense and is [in the docs as well](https://github.com/aasm/aasm#activerecord-enums).

In this specific case aasm should not redefine the question mark methods because the behaviour doesn't change from those already defined by ActiveRecord.

PR #341, opened 2016-03-15
Fix lazy error message messing up originating state

### LINKED ISSUE #312: InvalidTransition#message refers to machine's current state, not the state that caused the error
About this line: https://github.com/aasm/aasm/blob/master/lib/aasm/errors.rb#L13

We just had a day-long headache thinking over a weird error message we were getting from a pretty simple state machine (mostly truncated of unnecessary details):

```ruby
state :scheduled, :initial => true
state :running
state :error
state :success

event :retry do
  transitions to: :running, from: [:error]
end

event :fail do
  transitions to: :error, from: [:running, :scheduled, :error, :success]
end
```

We were getting the following in our error collection system:

    Event 'retry' cannot transition from 'error'

...and we were like **"wat?"** The state machine explicitly specified otherwise! Well... we actually had a bug in our app and fixed it. Here's what happened:

* We actually got an instance with a `scheduled` state. We attempted to `retry` it and got an `InvalidTransition`.
* Falling down into a `rescue` block, we `fail`ed the state machine (changing its state to `error`).
* We logged the exception into our collection system (`exception.message` and `exception.backtrace`)... except its message "specified the impossible" (given above) as it was referencing the machine's current state, `error`! Whereas when it happened the state was actually `scheduled`.

**Ignoring the shady programming practices one could detect above, is this the expected behavior? Shouldn't `InvalidTransition#message` indicate what the state was when it was raised, instead of using machine's current state?**

pirj:
Fixes #312

### DIFF
--- a/lib/aasm/errors.rb
+++ b/lib/aasm/errors.rb
@@ -3,18 +3,18 @@ module AASM
   class UnknownStateMachineError < RuntimeError; end
 
   class InvalidTransition < RuntimeError
-    attr_reader :object, :event_name, :state_machine_name, :failures
+    attr_reader :object, :event_name, :originating_state, :failures
 
     def initialize(object, event_name, state_machine_name, failures = [])
-      @object, @event_name, @state_machine_name, @failures = object, event_name, state_machine_name, failures
+      @object, @event_name, @originating_state, @failures = object, event_name, object.aasm(state_machine_name).current_state, failures
     end
 
     def message
-      "Event '#{event_name}' cannot transition from '#{object.aasm(state_machine_name).current_state}'. #{reasoning}"
+      "Event '#{event_name}' cannot transition from '#{originating_state}'. #{reasoning}"
     end
 
     def reasoning
-      "Failed callback(s): #{@failures}." unless failures.empty?
+      "Failed callback(s): #{failures}." unless failures.empty?
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/aasm/errors.rb
```diff
@@ -3,18 +3,18 @@ module AASM
...
     end
 
     def message
-      "Event '#{event_name}' cannot transition from '#{object.aasm(state_machine_name).current_state}'. #{reasoning}"
+      "Event '#{event_name}' cannot transition from '#{originating_state}'. #{reasoning}"
     end
 
     def reasoning
-      "Failed callback(s): #{@failures}." unless failures.empty?
+      "Failed callback(s): #{failures}." unless failures.empty?
```
alto: :+1:

lib/aasm/errors.rb
```diff
@@ -3,18 +3,18 @@ module AASM
   class UnknownStateMachineError < RuntimeError; end
 
   class InvalidTransition < RuntimeError
-    attr_reader :object, :event_name, :state_machine_name, :failures
+    attr_reader :object, :event_name, :originating_state, :failures
 
     def initialize(object, event_name, state_machine_name, failures = [])
-      @object, @event_name, @state_machine_name, @failures = object, event_name, state_machine_name, failures
+      @object, @event_name, @originating_state, @failures = object, event_name, object.aasm(state_machine_name).current_state, failures
```
alto: Indeed much better!

alto:
@pirj Whoa, what a nifty problem. Thanks for tracking down and improving!

alto:

PR #320, opened 2016-02-09
[RubyMotion support] Maintaining support after PR #316

Infotaku:
Here is the update to ensure RubyMotion support after merging PR https://github.com/aasm/aasm/pull/316.

PR #343, opened 2016-03-18
[RubyMotion support] Support for object proxies and removing the need for motion-bundler.

Infotaku:
Here is my first pass on providing support for proxied objects in RubyMotion. 

You already merged some of the code that should have been included here, I don't really know why one of my commits ended up in my previous pull request.

I'll try to come up with something a bit more automated in the next PR :wink:


Infotaku:
So, let me start by pointing out that most of the changes in my last commit may not be backward compatible with anything the user could have monkey-patched. It should however still be completely compatible with a more "by the book" usage.

I've worked around the "object proxy" issue by allowing AASM to make an educated guess about the class using it. If, when looking for a specific state machine, AASM finds that the class is not currently registered, it looks into its ancestors and returns the state machine for the first ancestor that includes AASM and has already been registered.

This is a big one, if you have any questions, feel free to ask.

alto:
@Infotaku Thanks for the PR. Could you take a look at the conflicts, please? Basically, what you say is this: with the latest merge of #320 _AASM_ doesn't work for RubyMotion anymore, and I should merge this PR as well to get that fixed?

Infotaku:
AASM still works with RubyMotion (even after PR #320), but as you can see in my [latest commit in this PR](https://github.com/aasm/aasm/pull/343/commits/3214e46f742085e35655617900d261daa5302bba), I refactored the way you store and retrieve the state machines, so anyone currently relying on an internal implementation of this would see their project break by updating the gem after this gets merged.

I'll look at the conflicts right away.

alto:

PR #346, opened 2016-04-04
Suppress warnings when using ActiveRecord enums feature

110y:
ActiveRecord enums feature defines the `state_name?` method, 
therefore `safely_define_method` warns `overriding method` whenever I use aasm with enums.
AASM works as expected, but it is little annoying for me.

So, this PR suppresses those warnings when using enums.
How is this?

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -193,7 +193,7 @@ module AASM
     end
 
     def safely_define_method(klass, method_name, method_definition)
-      if klass.instance_methods.include?(method_name.to_sym)
+      if !@options[:enum] && klass.instance_methods.include?(method_name.to_sym)
         warn "#{klass.name}: overriding method '#{method_name}'!"
       end
 


alto:
@110y I'd say that sounds pretty cool. Thanks a lot!

I just restarted the build on Travis. Let's see if that fixes the build errors.

alto:
Nope, that doesn't fix the build. I will have a look in to this.

alto:
Okay, fixed the problem. Happy to merge! 👍

110y:
@alto thx! 👍

leonelgalan:
Hey guys, @alto and @110y, the solution proposed here simply removes "disables" `safely_define_method` when `enum: true`. But we still want it for all other methods that weren't created because of `enum`. Maybe there is another way of asking "was the method previously defined by calling `enum`?" and disable the warning only on those instances.

holyketzer:
👍 
When it will be released?

alto:
@leonelgalan Please have a look at #353, which I think would solve your concern. Am I right?

alto:
@holyketzer Soon hopefully. I'll let you know :)

alto:
@110y, @holyketzer I just released version `4.10.1` including the PR.

holyketzer:
@alto thank you man, you saved all of us!

PR #353, opened 2016-04-18
Silence method override warning when it's an enum

nathanstitt:
This silences the `overriding method xxxx!` warning that's emitted when the overridden method came from an enum. 

I didn't notice #346 when I created this, but but this may be a better approach since it actually checks for defined enums using `defined_enums` (if present)

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -193,8 +193,8 @@ module AASM
     end
 
     def safely_define_method(klass, method_name, method_definition)
-      # Warn if method already exists and it did not originate from an enum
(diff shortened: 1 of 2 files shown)


alto:
@nathanstitt Hmm, I saw #346 first and merged it already. But I agree that your approach is more complete and would like to add it, too. Do you want to merge _master_ back into (and resolve the conflicts)? I'm happy to do it, just let me know 👍

nathanstitt:
Sure, just did so.  Thanks!

alto:
@nathanstitt Cheers!

alto:
@nathanstitt I just released version `4.10.1` including your PR.

PR #354, opened 2016-04-27
update_column instead of change_column

kikyous:


alto:
@kikyous Thanks a lot! 👍

PR #363, opened 2016-05-13
Fix a transition success callback issue

shunichi:
Transition success callback as Array got an error.


alto:
@shunichi Great spotting. I really appreciate it!

alto:
@shunichi I just released version `4.10.1` including your PR.

PR #364, opened 2016-05-14
Support :permitted => false for states and events

hspazio:
Given the following Job class
```ruby
class Job
  include AASM

  aasm do
    state :running, :cleaning

    event :run do
    end

    event :clean do
      transitions :from => :running, :to => :cleaning, :guard => :cleaning_needed?
    end

    event :sleep do
    end
  end
  
  def cleaning_needed?
    false
  end
end
```

and a `job` object currently in `:running` state
```ruby  
job.aasm.events(:permitted => false) #=> [:clean, :sleep]
# instead of the expected [:clean]

job.aasm.states(:permitted => false) #=> [:cleaning, :sleeping]
# instead of the expected [:cleaning]
```
This bug was caused by an ambiguous condition on `if options[:permitted]` which defaults to the full list of possible events/states if `:permitted => false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -866,29 +866,70 @@ end
 
 ### Inspection
 
-AASM supports a couple of methods to find out which states or events are provided or permitted.
+AASM supports query methods for states and events
 
-Given this `Job` class:
+Given the following `Job` class:
+```ruby
+class Job
+  include AASM
+
+  aasm do
+    state :sleeping, :initial => true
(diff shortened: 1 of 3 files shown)


alto:
@hspazio That's actually a very nice idea. I never thought of it like this. I really appreciate it!

alto:
@hspazio I just released version `4.10.1` including your PR.

Issue #262, opened 2015-09-30
bang methods don't raise validation errors leading to hard to trace errors


PragTob:
First of all, thanks for doing aasm :+1:

When a validation error occurs only in a state that I transition to, the `bang!` method returns false, I stay in the old state and I have no idea why the transition failed.

What I would like to have is to have some sort of access to the error, or even better a configuration option that the `bang!` methods raise on validation errors. Maybe that exists and I just don't know.

This way this was very hard to track down, why an object did not transition.

Now, taking this simple test class:

```ruby
class AasmTest < ActiveRecord::Base
  include AASM

  validates :number, numericality: true, if: :failing?

  aasm column: 'state' do
    state :init, intial: true
    state :failing

    event :bam do
      transitions from: :init, to: :failing
    end
  end
end
```

And now take it for a spin:

```
[2] pry(main)> a = AasmTest.new
=> #<AasmTest:0x0000000c6fcac8 id: nil, state: nil, number: nil>
[3] pry(main)> a.valid?
=> true
[4] pry(main)> a.bam!
   (0.1ms)  BEGIN
   (0.1ms)  COMMIT
=> false
[5] pry(main)> a.valid?
=> true
[6] pry(main)> # I have no idea why it did not transition as it is valid, there is no guard etc.
[7] pry(main)> a.bam
   (0.1ms)  BEGIN
   (0.1ms)  COMMIT
=> true
[8] pry(main)> a.valid?
=> false
[9] pry(main)> a.errors
=> #<ActiveModel::Errors:0x0000000c2d9838 @base=#<AasmTest:0x0000000c6fcac8 id: nil, state: "failing", number: nil>, @messages={:number=>["ist keine Zahl"]}>
[10] pry(main)> a.save!
   (0.1ms)  BEGIN
   (0.1ms)  ROLLBACK
ActiveRecord::RecordInvalid: Gültigkeitsprüfung ist fehlgeschlagen: Number ist keine Zahl
from /home/tobi/.rvm/gems/ruby-2.2.3@liefery-backend/gems/activerecord-4.2.3/lib/active_record/validations.rb:79:in `raise_record_invalid'
[11] pry(main)> # ah, that's why....
```

Thanks!

edit: I guess this somehow relates to #249 but I'm asking for a different thing :)

anilmaurya:
@PragTob This is not related to #249 . 
[...]

PR #366, opened 2016-05-16
Implement whiny persistence so that `transition!` will raise by default, mirroring ActiveRecord's behavior.

Wildebeest:
@PragTob's suggestion from https://github.com/aasm/aasm/issues/262.

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -76,7 +76,12 @@ module AASM
             self.save
           end
 
-          success ? true : aasm_rollback(name, old_value)
+          unless success
+            aasm_rollback(name, old_value)
+            raise ActiveRecord::RecordInvalid.new(self) if aasm_whiny_persistence(name)
+          end
+
+          success
         end
 
         # Writes <tt>state</tt> to the state column, but does not persist it to the database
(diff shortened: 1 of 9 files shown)


PragTob:
Thanks for taking a stab at this and LGTM. :) Not sure about the version bump though, should be done for semantic versioning reasons but in the end maintainer's call :)

Wildebeest:
Definitely. I can also take out the version bump or change it to whatever @alto wants.

lib/aasm/persistence/active_record_persistence.rb
```diff
@@ -76,7 +76,12 @@ def aasm_write_state(state, name=:default)
             self.save
           end
 
-          success ? true : aasm_rollback(name, old_value)
+          unless success
+            aasm_rollback(name, old_value)
+            raise ActiveRecord::RecordInvalid.new(self) if aasm_whiny_persistence(name)
+          end
+
+          success
```
tessi: Since the only possible value for `success` at this point is `true`, I'd be explicit and would return `true` here instead of `success` .
Might be a personal thing, though.

Wildebeest: It can return `false` if `aasm_whiny_persistence` is false.

tessi: Ah, I see. Nvermind then :)

alto:
@Wildebeest Thanks a lot for your pull request, great work! 

I even support the default setting for `whiny_persistence` to be true, although that indeed requires a new major version. Let's do this: I accept your code changes as they are, make `whiny_persistence: false` the default and release it as new minor version. I will then change the default to `true` as part of the upcoming release `5.0.0`. What do you think?

alto:
@Wildebeest And by the way, sorry coming back to this so late!

Wildebeest:
@alto No worries! I'm glad you were able to take a look. :)

Works for me. I've rebased and updated the PR with your suggestions.

alto:
@Wildebeest  I just released version `4.11.0` including your change. 👍

alto:
@Wildebeest Have a look at #378, which will be part of the upcoming release `5.0.0`.

Wildebeest:
<3 Thanks @alto!

PR #367, opened 2016-05-24
Add :after_commit callbacks to events.

rringler:
This PR reimplements @hjdivad's [patch](https://github.com/aasm/aasm/pull/170) delegating event `after_commit` callbacks 
when using ActiveRecord.  This allows the `after_commit` callbacks to be fired 
both when the record is persisted immediately (eg., via `#event!` methods) as  well 
as when the record is persisted manually (eg., via the `#event` method, followed by 
a `#save`.)  It also preserves the order of `after_commit` callbacks in the case of 
multiple state transitions.

This (patched) behavior is  more  consistent with how ActiveRecord handles 
`after_commit` callbacks.

Edit: I apologize for creating a new PR for this; I've lost access to the repo from
which the first PR originated from.

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -33,6 +33,8 @@ module AASM
 
         base.after_initialize :aasm_ensure_initial_state
 
+        base.after_commit :aasm_after_commit_hooks
+
         # ensure state is in the list of states
         base.validate :aasm_validate_states
       end
@@ -189,8 +191,10 @@ module AASM
             end
 
             if options[:persist] && success
-              event.fire_callbacks(:after_commit, self, *args)
-              event.fire_global_callbacks(:after_all_commits, self, *args)
+              # Delegating to ActiveRecord
+              #
+              # event.fire_callbacks(:after_commit, self, *args)
+              # event.fire_global_callbacks(:after_all_commits, self, *args)
             end
           ensure
             if options[:persist]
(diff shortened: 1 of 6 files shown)


lib/aasm/persistence/active_record_persistence.rb
```diff
@@ -189,8 +191,10 @@ def aasm_fire_event(state_machine_name, name, options, *args, &block)
             end
 
             if options[:persist] && success
-              event.fire_callbacks(:after_commit, self, *args)
-              event.fire_global_callbacks(:after_all_commits, self, *args)
+              # Delegating to ActiveRecord
```
HoyaBoya: Super neat stuff. In my humble opinion, perhaps it's just better to remove the lines below vs comment out.

hjdivad:
@rringler should probably close #298 with a note that this PR supersedes it.

lib/aasm/persistence/active_record_persistence.rb
hjdivad: It looks like this block isn't even meant to be changed at all.  These lines look exactly like the ones this patch deletes, above.

alto:
@rringler Thanks for the new PR. I really appreciate your help with this! 👍 

I resolved the actual conflicts and merged this PR into the _aasm5_ branch.

alto:
I will let you know once _AASM_ version 5 is released.

rringler:
Hey @alto.  Thank you for finishing this off.  I apologize for letting it sit for a while; I've been busy transitioning at work.

swrobel:
Did this end up being released with 5.0? I don't actually see it anywhere in the CHANGELOG.

swrobel:
@anilmaurya you seem to be the new maintainer - do you know the answer here?

axsuul:
@swrobel It doesn't look like it was merged. Have you found a workaround?

swrobel:
> Have you found a workaround?

Not really, I'm just only using the bang (`!`) event methods but it feels brittle and like someone could introduce a bug at any moment until this gets fixed...

axsuul:
Gotcha, thanks for chiming in!

Issue #361, opened 2016-05-12
Config option to disable override warnings


t3hk0d3:
Method override warnings were remade in 4.10.0 and brought some painful experience:

```
MySchema: overriding method 'active?'!
MySchema: overriding method 'subscribe'!
MySchema: overriding method 'unsubscribe'!
MySchema: overriding method 'activate'!
MySchema: overriding method 'suspend'!
MySchema: overriding method 'cancel'!
```

Generally its good idea to inform developers about possible problems, but in my case it were made intentionally.

It would be a great thing to give developers opportunity to disable these warning globally or per-state machine, since they have great impact on developer psychic health, especially if you have plenty state machines in project over same abstract class. Specifying ``-W0`` is not an option.

For example:
```ruby
AASM.config.disable_override_warning = true

aasm(disable_override_warning: true) do
...
end
```

Thank you!

alextaylor000:
This looks like it could be at least partially related to #346. I'm using AASM with an enum column, and I found I was getting "overriding method" errors for a bunch of question-mark methods until I explicitly set `enum: true`:

```ruby
aasm column: :state, enum: true do
  # ...
end
```

t3hk0d3:
@alextaylor000 I am not using AASM in ActiveRecord class at all. ``enum: true`` won't help me.

bacala:
You can hide warnings by setting AASM::Configuration.hide_warnings = true

For me it worked just fine

PR #369, opened 2016-05-25
Allow the ability to ignore warnings on intentional overriden methods

HoyaBoya:
This PR introduces an ability to ignore or acknowledge an override attempt when AASM defines methods.

I've read over the source PR from @pirj and it makes a lot of sense. Perhaps it's a reflection of an AASM anti-pattern that we are using over here, but we tend to create a lot of Base / Parent classes that we eventually implement with AASM sub classes. 

The quasi Ruby `abstract` pattern looks like the following:

```
class SomeParent
  def drink_milk
    raise NotImplementedError, "Don't need to drink anymore!"
  end
end

class SomePunkKid < SomeParent
  aasm do
    state :eating_cookies
    state :milk_time

    event :drink_milk do
      transitions from: :eating_cookies, to: :milk_time, before: :pour_glass, after: :drink_some_milk
    end
  end

  def pour_glass
    # pour
  end

  def drink_some_milk
    # gulp
  end
end
```

It's a slight burden to now get warnings via `Kernel#warn` that we cannot silence with the appropriate Ruby arguments.

This PR will offer two override configurations to handle this. The first, and most likely to be used, is a white list of `override_methods` that one can add to suppress a warning. There's also a more dangerous `ignore_override_warnings` that kills all messaging. I introduced this when trying to silence the warnings in the test suite where it was becoming tedious to enumerate all of the methods that were intentionally overridden.

Overall, I'm not 100% sure on this. There's an argument to be made that we shouldn't do the warning and that proper test coverage should catch these issues. I've also thought about making a logger configuration so we can not send to Kernel. Unsure.

Let's digest this for a bit, and if we feel the direction is worthy, I'll be happy to provide a touch commit with documentation.

Related Issues
https://github.com/aasm/aasm/issues/361

Related PR
https://github.com/aasm/aasm/pull/340

### DIFF
--- a/lib/aasm/configuration.rb
+++ b/lib/aasm/configuration.rb
@@ -34,5 +34,11 @@ module AASM
 
     # Configure a logger, with default being a Logger to STDERR
     attr_accessor :logger
+
+    # These are the methods that will be override in a sub-class.
+    attr_accessor :override_methods
+
+    # Ignore all warnings, to be used at your own risk.
+    attr_accessor :ignore_override_warnings
   end
 end
--- a/lib/aasm/configuration.rb
+++ b/lib/aasm/configuration.rb
@@ -37,8 +37,5 @@ module AASM
 
     # These are the methods that will be override in a sub-class.
     attr_accessor :override_methods
-
-    # Ignore all warnings, to be used at your own risk.
-    attr_accessor :ignore_override_warnings
   end
 end
--- a/README.md
+++ b/README.md
@@ -407,8 +407,14 @@ _AASM_ doesn't prohibit to define the same event in more than one state machine.
 latest definition "wins" and overrides previous definitions. Nonetheless, a warning is issued:
 `SimpleMultipleExample: overriding method 'run'!`.
 
+If you want to acknowledge the override to suppress the warning, you may do so with `override_methods`:
+
+```ruby
+aasm(:move, override_methods: [:run!])
+```
+
 All _AASM_ class- and instance-level `aasm` methods accept a state machine selector.
-So, for example, to use inspection on a class level, you have to use
+So, for example, to use inspection on a class level, you have to use:
 
 ```ruby
 SimpleMultipleExample.aasm(:work).states
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -210,16 +210,10 @@ module AASM
              klass.respond_to?(:defined_enums) &&
              klass.defined_enums.values.any?{ |methods|
                  methods.keys{| enum | enum + '?' == method_name }
-             })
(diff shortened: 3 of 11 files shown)


lib/aasm/configuration.rb
```diff
@@ -28,5 +28,11 @@ class Configuration
 
     # namespace reader methods and constants
     attr_accessor :namespace
+
+    # These are the methods that will be override in a sub-class.
+    attr_accessor :override_methods
+
+    # Ignore all warnings, to be used at your own risk.
+    attr_accessor :ignore_override_warnings
```
pirj: I'm really afraid that this this will become a mainstream aasm definition and people will still be overriding methods silently.
I would go even further and would raise an exception (a cumulative one after all definitions are compiled) in development environment to force to either ignore the warnings or to fix the names of the state/transition/concurrent property/method.

pirj:
👏 
Needs explanation in readme.

lib/aasm/configuration.rb
HoyaBoya: Perhaps we even get rid of it and not even allow a slippery slope here?

Let me get a touch commit and see what people think

HoyaBoya:
@pirj feedback incorporated on removing `ignore_override_warnings` and adding documentation.

pirj:
@HoyaBoya awesome!

t3hk0d3:
I'd suggest option to disable override warning for whole event machine.

```ruby
aasm(override_warnings: false) do
  ...
end
```

pirj:
@t3hk0d3 Related https://github.com/aasm/aasm/pull/369#discussion_r64632735

HoyaBoya:
@alto, do you want to weigh in on the direction you'd like to see this go? I see both sides of the equation and happy to accommodate whatever the community feels is best.

t3hk0d3:
@pirj 
You are too pessimistic about our community. People aren't that bad as you think.

HoyaBoya:
@alto @pirj @t3hk0d3, I've rebased master into this branch so that we log the warnings to the configuration logger. @alto do you want to provide some guidance on whether we want a general `:ignore_override_warnings` configuration or not?

Again I see both sides, but if I had to choose, I'd argue on behalf of @pirj's point of view and ease into this slowly. First with an explicit set of white list values that a developer sets, and we see how this digests. Then take it from there?

https://github.com/aasm/aasm/pull/369/commits/bc358f23e1a3aebfea0c2339ffda5a5abbd138d9

HoyaBoya:
Should I close this and instead defer to #370?

I guess if the message vexes others, one can just pipe to `/dev/null` via

```
aasm logger: Logger.new('/dev/null') do

end
```

PR #370, opened 2016-05-25
Allow AASM to Take A Logger Configuration

HoyaBoya:
This is tangentially related to #369. In this PR, we're creating the ability to specify the AASM logger to warn against when a method is overridden. It currently logs to `Kernel#warn`.

http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-warn

This is is created as a separate PR in the event we feel that #369 is not the direction we want. Perhaps a combination of both might offer us greater flexibility, or perhaps not.

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -1,3 +1,5 @@
+require 'logger'
+
 module AASM
   class Base
 
@@ -39,6 +41,9 @@ module AASM
 
       # Set to true to namespace reader methods and constants
       configure :namespace, false
+
+      # Configure a logger, with default being a Logger to STDERR
+      configure :logger, Logger.new(STDERR)
 
       # make sure to raise an error if no_direct_assignment is enabled
       # and attribute is directly assigned though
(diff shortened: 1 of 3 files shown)


lib/aasm/base.rb
```diff
@@ -40,6 +42,9 @@ def initialize(klass, name, state_machine, options={}, &block)
       # Set to true to namespace reader methods and constants
       configure :namespace, false
 
+      # Configure a logger, with default being a Logger to STDERR
+      configure :logger, Logger.new(STDERR)
```
HoyaBoya: Quick question here, do we feel that we might want to introduce a `AASM::Configuration` that can allow for global changes across all `aasm` blocks? Not sure if this is a step too far for this PR or where the direction AASM wants to go. Not even sure how the precedence of logger configurations are honored, (local => global => Kernel)

spec/unit/override_warning_spec.rb
```diff
@@ -28,42 +28,59 @@ def defined_enums
   end
 
   describe 'state' do
-    class Base
-      def valid?; end
+    let(:base_klass) do
```
HoyaBoya: I did this to prevent class bleed across specs.

alto: Nice!

lib/aasm/base.rb
alto: I love small pull requests. So yes, it would be too much for this one. But I'm happy to see another pull request for your `AASM::Configuration` approach (there is already a class like this, which might be enhanced for this purpose).

alto:
@HoyaBoya Great pull request. Thanks a lot!

alto:
@HoyaBoya I just released version `4.11.0` including your change. 👍

PR #378, opened 2016-06-19
Let whiny_persistence: true be the default now

alto:
Based on #366, this PR sets the default `whiny_persistence` setting to true (DSL/behavior change). So, by default, `transition!` will fail if the object is invalid.
- [x] change default to `true`
- [ ] document configuration option


andyw8:
I'm confused, the changelog says for v3.0.0:

> whiny transactions: by default, raise an exception if an event transition is not possible

Which seems to not be the case, since this PR is open?

swrobel:
@andyw8 `whiny_transitions` != `whiny_persistence` https://github.com/aasm/aasm/blob/v4.12.3/lib/aasm/base.rb#L18

swrobel:
@alto any reason why 5.0 was released without this? Also, I'm curious why the bump to 5.x if it seems like there weren't breaking changes...

anilmaurya:
@swrobel We bumped to 5.x because of internal architecture change in AASM

AlexanderZagaynov:
Is there a way to set in with the current version by default on app load/init stage?

route:
👍🏻  should be by default `true`

danielvdao:
@anilmaurya is there any reason this shouldn't be merged?

afuno:
Hello. Is there any information on when this change will be released?

PR #379, opened 2016-06-28
After and after_all_transitions_callback execution order

Ashviniv:
`after transition` should be called before `after_all_transitions`, like after event is called before after_all_events.But currently, `after` transition is called after `after_all_transitions`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,8 +175,8 @@ begin
   transition      guards
   old_state       before_exit
   old_state       exit
-                  after_all_transitions
   transition      after
+                  after_all_transitions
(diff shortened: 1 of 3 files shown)


joallard:
On the failing tests: Looks like the return value from the `:after` callback should be tapped into to be returned on transition.

I arrived here after a search because I agree with the spirit, but this is probably a breaking-change.

anilmaurya:
@Ashviniv I think you have a valid point, could you rebase your commit with master branch and resolve conflicts.

As this is breaking-change we can release this change in AASM 5.0

PR #380, opened 2016-07-05
Update README.md

danlo:
Removed (user) which I think should not be present.  There is no argument passed to the method call.


alto:
Cheers! 👍

Issue #383, opened 2016-07-14
Failures array in transition not being reset.


jmcnevin:
I'm wondering if anyone else has run into this issue with AASM 4.11 and Rspec.  I'm not sure if it relates to using `let` to define variables in specs.

As my specs run, I start to see failures in state transitions that worked in AASM 4.7, and the list of failures seems to compound with each spec run.  Example:

```
Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?].
Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?].
Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?].
Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?, :ready_to_close?].
```

I have a feeling this is related to storing the failures in an instance variable on the transition object, and somehow that array of failures isn't being cleared properly between transition attempts.  Is there a way I should be configuring my test suite to prevent this?

jmcnevin:
Still having problems with this.

jmcnevin:
Attempted to fix this in my PR, but I'm still not following what's happening here:

```
x.may_close?
=> true
x.close!
=> *** AASM::InvalidTransition Exception: Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?].
x.may_close?
=> false
```

It appears some state is still being persisted between transition attempts/queries, but I'm having a tough time tracking down where this is happening.

adamniedzielski:
I managed to reproduce the bug against AASM 4.12.3 and narrow it down to a simple example. It's not specific to tests, we run into this bug in our production code for reporting exceptions from state transitions.

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'activerecord', '5.1.4'
  gem 'sqlite3'
  gem 'aasm', '4.12.3'
end

require 'active_record'
require 'sqlite3'
require 'aasm'

puts "AASM: #{AASM::VERSION}"

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')

[...]

PR #384, opened 2016-07-14
Clear transition failures between attempts.

jmcnevin:
This is related to issue #383.

The failures array is not being cleared between transition attempts, preventing an object from transitioning even if the guards would normally permit it.

A few specs needed to be modified, as they were expecting this behavior.

### DIFF
--- a/lib/aasm/core/transition.rb
+++ b/lib/aasm/core/transition.rb
@@ -29,11 +29,13 @@ module AASM::Core
     end
 
     def allowed?(obj, *args)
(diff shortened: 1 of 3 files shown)


jmcnevin:
Still trying to understand how I can get into a situation like this:

```
x.may_close?
=> true
x.close!
=> *** AASM::InvalidTransition Exception: Event 'close' cannot transition from 'available'. Failed callback(s): [:ready_to_close?, :ready_to_close?].
x.may_close?
=> false
```

PR #386, opened 2016-07-19
Added documentation for auto-generated constants

dakotahp:
This is a great feature that I’ve researched in the past and found no information before. A colleague (@aguynamedloren) finally found a merged PR related to this and it’s a shame it isn’t documented.


anilmaurya:
@dakotahp Thank you ♡

Issue #388, opened 2016-07-27
Permissible states not respecting guard parameters


froske:
I understand calling for permissible states works with guards (ref issue #54), but I can't get it to work when my guard has a parameter.  It seems like it's not getting passed to the guard (only when I call for permissible states.

Example:

```
state :new, :reviewed, :finalized

event :mark_as_reviewed do
  transitions :from => :new, :to => :reviewed, :guards => [:user_is_manager?]
end

...

def user_is_manager?(usr)
  ok = false
  if usr.has_role? :manager
    ok = true
  end
  return ok
end
```

```
$ u = User.find(100)
$ u.has_role? :manager
=> true

$item.aasm.current_state
=> :new

$ item.may_mark_as_reviewed?(u)
=> true

$ item.aasm.states(:permitted => true, :usr => u).map(&:name)
=> NoMethodError: undefined method `has_role?' for nil:NilClass.....

```

but the guard works in other instances with a parameter, like when I actually try to change the state.

```
$ item.mark_as_reviewed!(u)
=> true

$ item.aasm.current_state
=> :reviewed
```

If I do a `puts usr.inspect` inside the guard method, it shows me that the User object did in fact make it when I called `item.may_mark_reviewed?(u)`  but when I call for permitted states it's just nil.

I've tried every syntax I can think of. 

What am I missing?

anilmaurya:
Fixed with #389

anilmaurya:
@froske Thank you for reporting this bug, its fixed now in version `4.11.1 `.

PR #389, opened 2016-07-31
Fix [#388] guard parameter for permissible states

anilmaurya:


anilmaurya:
@alto  Can you check why bundle install is failing for mongo_mapper.gemfile on travis ci

PR #390, opened 2016-08-01
The dependency `tins` 1.7.0 officially require ruby >= 2.0

anilmaurya:
- https://github.com/sferik/rails_admin/issues/2464


anilmaurya:
Didn't solved tins dependency issue, therefore closing it.

alto:
@anilmaurya I think the problem comes from `coveralls` which – for unclear reasons – was still part of some of the gemfiles used for the Travis CI build. Should be fixed now on _master_.

anilmaurya:
Yes, failing specs are now passing on travis 🏆

PR #398, opened 2016-09-19
added custom-name to generator to avoid uninitialized constant error

bastianwegge:
I added the column name to the generator including an additional spec for future use. Let me know if there are any downsides to my changes.


anilmaurya:
@bastianwegge Thank you for fixing this issue. I don't think there is any downside with your changes. 👍

anilmaurya:
@bastianwegge I released `4.11.1` including your changes.

bastianwegge:
Thanks for the Info and have a nice day!

PR #406, opened 2016-10-04
Ensure nil arg is forwarded to callbacks

### LINKED ISSUE #404: Argument is not passed to the callback if it is nil
Given following,

``` ruby
aasm do
  event :example do
    transitions from: :t1, to: :t2, after: :do_stuff
  end
end

def do_stuff(payload)
end
```

If I do `object.example!("Hello, world!")`, it works. However, if I do `object.example!(nil)`, it returns an `ArgumentError: wrong number of arguments (given 0, expected 1)` for `do_stuff` method.

yogeshjain999:
fixes #404

### DIFF
--- a/lib/aasm/core/event.rb
+++ b/lib/aasm/core/event.rb
@@ -122,6 +122,8 @@ module AASM::Core
           args.unshift(to_state)
           to_state = nil
         end
+      else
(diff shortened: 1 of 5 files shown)


anilmaurya:
Thank you @yogeshjain999 for helping out . :sparkles:

Issue #372, opened 2016-05-27
Multiple Transitions with guards and after execute wrong callback


lamtha:
event :go  do
  transitions from: [:a,:b], to: :c, :guard => :returnsfalse, after: :this_gets_executed
  transitions from: [:d,:e], to: :c, :guard => :returnstrue,  after: :this_doesnt_execute
end

Removing the second transition raises:

InvalidTransition Failed callback(s): [:returnsfalse]

asmega:
are you saying this code raises the exception?
```
event :go do
  transitions from: [:a,:b], to: :c, :guard => :returnsfalse, after: :this_gets_executed
end
```

lamtha:
Yes

asmega:
your guard is returning false which prevents the transition from happening. there are no other possible transitions hence it raises the error. this is expected behaviour

lamtha:
Agreed, for the code you created, the behavior is correct.
 
For my sample code, the wrong after method gets executed (the state transition works).

asmega:
below is a full example on the latest version

```
[1] pry(main)> class Job
[1] pry(main)*   include AASM
[1] pry(main)*
[1] pry(main)*   aasm do
[1] pry(main)*     state :a, :initial => true
[1] pry(main)*     state :b
[1] pry(main)*     state :c
[1] pry(main)*
[1] pry(main)*     event :go do
[1] pry(main)*       transitions from: [:a,:b], to: :c, :guard => :returnsfalse, after: :this_gets_executed
[1] pry(main)*       # transitions from: [:d,:e], to: :c, :guard => :returnstrue, after: :this_doesnt_execute
[1] pry(main)*     end
[1] pry(main)*   end
[1] pry(main)*
[1] pry(main)*   def returnsfalse
[1] pry(main)*     false
[1] pry(main)*   end
[1] pry(main)*
[1] pry(main)*   def returnstrue
[1] pry(main)*     true
[1] pry(main)*   end
[1] pry(main)*
[1] pry(main)*   def this_gets_executed
[1] pry(main)*     raise :this_gets_executed
[1] pry(main)*   end
[1] pry(main)*
[1] pry(main)*   def this_doesnt_execute
[1] pry(main)*     raise :this_doesnt_execute
[1] pry(main)*   end
[1] pry(main)* end
=> :this_doesnt_execute
[2] pry(main)>
[3] pry(main)> job = Job.new
=> #<Job:0x007fab7d3a8588>
[4] pry(main)> job.go
[...]

Issue #362, opened 2016-05-12
Transitions with same src and dst states do not respect guards


t3hk0d3:
Having two transitions in same event with same source and destination states, but different guards cause unexpected behavior (called first transition with matching src/dest states, disregarded guards).

Example:

```ruby
class MyModel
  include AASM
  aasm do
    state :one, initial: true
    state :two

    event :start do
      transitions from: :one, to: :two, if: [:abc?], after: [:do_foo]
      transitions from: :one, to: :two, unless: [:abc?], after: [:do_boo]
    end
  end

  def abc?
    # compute something
    false # for example
  end

  def do_foo
    puts  "FOO"
  end

  def do_boo
    puts "BOO"
  end
end

obj = MyModel.new

obj.abc?
# => false

obj.start! 
# => FOO # but expected BOO, since abc? => false
```

Current workaround is to move guard logic into callback:

```ruby
transitions from: :one, to: :two, after: [:do_thing]
...
def do_thing
   if abc?
      puts "FOO"
   else
      puts "BOO"
   end
end
```

Tongboy:
+1 - just started experiencing this today with 4.11.1

anilmaurya:
Thank you @t3hk0d3 for reporting this issue, its now resolved in master branch.

PR #408, opened 2016-10-12
Allow multiple transitions in a single event with the same to and from states

dathanb:
As reported in https://github.com/aasm/aasm/issues/372 and https://github.com/aasm/aasm/issues/362, if two transitions in the same event have the same `to` and `from` states but differ in their guards, if any transition is valid, the first transition is the one that's always executed. This PR attempts to resolve that issue.


anilmaurya:
@dathanb Thank you for your contribution ♡

PR #412, opened 2016-10-20
Update README.md - Active Record

mateuszwieloch:
Changes in #366 cause ActiveRecord persistence to throw exceptions if `whiny_persistence` flag is set. Updating README to reflect that change.


anilmaurya:
@mateuszwieloch Thank you for your contribution 👍

PR #422, opened 2016-12-05
Add before_success callback on event layer

timsly:
Sometimes you need to call some global callback on event **before** some transition callback but **after** state was persisted  successful. I didn't find such callback in the current codebase so I introduced `before_success` callback.

Here is some simple case where it can be useful
```ruby
event :send do
  before_success do
    # update object here, to make sure we notify recipient or sender with up-to-date object
    update! color: 'red'
  end

  transition from: :recipient_turn, to: :sender_turn do
    success { notify :sender }
  end

  transition from: :sender_turn, to: :recipient_turn do
    success { notify :recipient }
  end
end
```

### DIFF
--- a/spec/unit/callbacks_spec.rb
+++ b/spec/unit/callbacks_spec.rb
@@ -118,6 +118,7 @@ describe 'callbacks for the new DSL' do
       expect(callback).to receive(:before_enter_closed).once.ordered
       expect(callback).to receive(:enter_closed).once.ordered
       expect(callback).to receive(:aasm_write_state).once.ordered.and_return(true)   # this is when the state changes
+      expect(callback).to receive(:event_before_success).once.ordered
       expect(callback).to receive(:success_transition).once.ordered.and_return(true) # these should be after the state changes
       expect(callback).to receive(:after_exit_open).once.ordered
       expect(callback).to receive(:after_enter_closed).once.ordered
@@ -143,6 +144,7 @@ describe 'callbacks for the new DSL' do
     expect(callback).to_not receive(:before_enter_closed)
     expect(callback).to_not receive(:enter_closed)
     expect(callback).to_not receive(:aasm_write_state)
+    expect(callback).to_not receive(:event_before_success)
     expect(callback).to_not receive(:success_transition)
     expect(callback).to_not receive(:after_exit_open)
     expect(callback).to_not receive(:after_enter_closed)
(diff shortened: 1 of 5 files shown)


anilmaurya:
@timsly Thank you for your effort but I didn't understood use case for `before_success` callback .

In your example if I use `before` callback then how it will make a difference ?

timsly:
@anilmaurya `before` happen **before** state was persisted and `before_success` happen **after** state was presisted, but before `success`

anilmaurya:
I understand that but is it going to make any difference in result ? 

If I do 

```
event :send do
  transition from: :recipient_turn, to: :sender_turn do
    after :update_color
    success { notify :sender }
  end

  transition from: :sender_turn, to: :recipient_turn do
    after :update_color
    success { notify :recipient }
  end
end

def update_color
   update! color: 'red'
 end

```

Will this solve the problem you are solving with `before_success` callback

timsly:
No, it won't. Because `after` for transition still happen **before** state was presisted.
Theoretically we can do this with
```ruby
event :send do
  transition from: :recipient_turn, to: :sender_turn do
    success { update_color; notify :sender }
  end

  transition from: :sender_turn, to: :recipient_turn do
    success { update_color; notify :sender }
  end
end

def update_color
  update! color: 'red'
end
```

But I think it's better to have some callback that we can use for such kind of cases

anilmaurya:
Sorry to trouble you even more, please bear with me.

You goal is to update object(make color red), to make sure notify recipient or sender get up-to-date object.

How does it matter if I update object before state is persisted or not.

timsly:
You are not troubling me, I am happy that we can discuss and, hope, solve that :-)

Example with color is synthetic. In my app I calculate message based on the current state and then notify. So it's important to have state up-to-date there.
Probably I used wrong word to describe that I am not worrying about persistence of the state but it's important to have state up-to-date but `after` callback for transition happen before that(state was updated)

anilmaurya:
Ok, Understood your point.

Will get back to you soon after reviewing your pull request.

timsly:
Cool. Thanks @anilmaurya

spec/unit/callbacks_spec.rb
```diff
@@ -118,6 +118,7 @@
       expect(callback).to receive(:before_enter_closed).once.ordered
       expect(callback).to receive(:enter_closed).once.ordered
       expect(callback).to receive(:aasm_write_state).once.ordered.and_return(true)   # this is when the state changes
+      expect(callback).to_not receive(:event_before_success).once.ordered
```
anilmaurya: It should be
```
expect(callback).to receive(:event_before_success).once.ordered
```

timsly:
@anilmaurya `callbacks_spec.rb` file was updated. Also, I didn't find anything about squash policy for PR so I squashed my commits.

anilmaurya:
Thank you @timsly for squashing commits and for your contribution 🏄🏻

PR #426, opened 2016-12-23
Removed dead code

hlegius:
Just a small 💅  👯

### DIFF
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -14,7 +14,6 @@ module AASM
 
     def current_state=(state)
       @instance.aasm_write_state_without_persistence(state, @name)
-      # @current_state = state
     end
 
     def enter_initial_state
@@ -22,7 +21,6 @@ module AASM
       state_object = state_object_for_name(state_name)
 
       state_object.fire_callbacks(:before_enter, @instance)
-      # state_object.fire_callbacks(:enter, @instance)
       self.current_state = state_name
       state_object.fire_callbacks(:after_enter, @instance)
 


lib/aasm/instance_base.rb
```diff
@@ -14,15 +14,13 @@ def current_state
...
       @instance.aasm_write_state_without_persistence(state, @name)
-      # @current_state = state
     end
 
     def enter_initial_state
       state_name = determine_state_name(@instance.class.aasm(@name).initial_state)
       state_object = state_object_for_name(state_name)
 
       state_object.fire_callbacks(:before_enter, @instance)
-      # state_object.fire_callbacks(:enter, @instance)
```
anilmaurya: @alto Did you planned to remove `enter` callback for state ? 

I can see we are still calling `enter` callback https://github.com/aasm/aasm/blob/master/lib/aasm/aasm.rb#L145 with a TODO: remove for AASM 4.

Should we remove `enter` callback from AASM ?

alto: @anilmaurya When I added this TODO, I wasn't sure if we still need/want to support this callback. Reason is that actually there is no difference between the `enter` and the `before_enter` callback (there is nothing happening in between). It's just a semantic difference (redundant).

What do you think?

As this would be a DSL change, I initially wanted to make that change for _AASM 4_.

anilmaurya: It make sense to remove redundant callbacks, I think we should make this change in AASM 5.

anilmaurya:
Merging this after adding backlog card https://github.com/aasm/aasm/projects/1#card-1456801

anilmaurya:
@hlegius Thank you .

Issue #418, opened 2016-11-22
Shortcut for timestamps


Drenmi:
A common and recurring thing we do is implement callbacks for updating timestamps. How about an option for `state` that automatically updates the timestamp when the record enters that state?

```ruby
state :finished, timestamp: true
# Will look for an attribute `finished_at`
```

perhaps with an optional attribute name for flexibility:

```ruby
state :draft, timestamp: :drafted_at
```

WDYT?

anilmaurya:
I liked optional attribute name flexibility.

Would like to have this feature in AASM, are you interested in sending a PR ?

CapCap:
@anilmaurya am getting the same thing as #424

```
Gem::InstallError: nokogiri requires Ruby version >= 2.1.0.
Installing aws-sdk-core 2.6.44
An error occurred while installing nokogiri (1.7.0.1), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.7.0.1'` succeeds before bundling.
```

Drenmi:
@anilmaurya: I might be able to submit a PR, once I have some spare time. 😅 If anyone comes across this and wants to work on it, feel free. 😊

anilmaurya:
@CapCap At first it look looks like `nokogiri 1.7.0.1` dependency on `ruby >= 2.1.0` may be the reason behind failing specs. I will debug further to solve it.
@Drenmi Looking forward for pull request 😊

anilmaurya:
@Drenmi Just came across #427 , @CapCap is working on this.

lichtamberg:
Can be closed after this was merged or?

Drenmi:
@lichtamberg: The relevant PR (#427) is still open. I don't think anyone is working on it though. 🙂

kbighorse:
Was the OP's mentioned optional attribute name for flexibility ever implemented?

`state :draft, timestamp: :drafted_at`

PR #427, opened 2017-01-08
automatically set timestamps on state update #418

CapCap:
an attempt at #418

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -34,6 +34,9 @@ module AASM
       # string for a specific lock type i.e. FOR UPDATE NOWAIT
       configure :requires_lock, false
 
+      # Set to true to automatically set `"#{state_name}_at" = ::Date.now` on state changes
+      configure :timestamp, false
+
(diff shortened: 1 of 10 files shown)


reidmorrison:
Rather than building the `timestamp` capability directly into every persistence layer supported by AASM, could the `timestamp` mechanism generate a callback under the covers that calls the setter for the column?

This will also keep more advanced logic out of every persistence layer so that they can focus purely on persistence.

By enabling the global timestamp config, will it break any gems that the application uses that also use AASM, but do not have any timestamps?

anilmaurya:
@reidmorrison comment sounds good to me, @CapCap @Drenmi What do you guys think ?

Drenmi:
Sounds like a sensible approach @reidmorrison. 🙂

CapCap:
@reidmorrison I like it too. Not sure how much time I have to do it though, but will try

pynixwang:
what callback should I use to set timestamp column?

pynixwang:
before is ok

PR #429, opened 2017-01-10
Deep clone state machine during inheritance (with tests)

Tybot204:
I need the ability to modify state machines in child classes in a project I'm currently working on, and came across this PR: https://github.com/aasm/aasm/pull/269

I have added extra tests to answer @alto's concerns and rebased with the current master. Let me know if anything else is needed!

### DIFF
--- a/lib/aasm/core/event.rb
+++ b/lib/aasm/core/event.rb
@@ -27,6 +27,16 @@ module AASM::Core
       ], &block) if block
     end
 
+    # called internally by Ruby 1.9 after clone()
+    def initialize_copy(orig)
+      super
+      @transitions = @transitions.collect { |transition| transition.clone }
+      @guards      = @guards.dup
+      @unless      = @unless.dup
+      @options     = {}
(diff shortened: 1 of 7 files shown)


Tybot204:
Failing builds on Travis CI are related to old Ruby versions and gems installed here. Should have nothing to do with this PR.

Tybot204:
After playing around with this in my project, it seems the `@global_callbacks` instance variable is not correctly getting duplicated. If you load both the superclass and subclass, it will contain both sets of global callbacks. Looking into this now...

reidmorrison:
:+1: Nice tests! :tada:

reidmorrison:
@anilmaurya please accept this PR. If there are any concerns, we automatically include these patches in another open source gem [Rocket Job](https://github.com/rocketjob/rocketjob) which makes extensive use of inheritance with AASM.

To see the patches: https://github.com/rocketjob/rocketjob/blob/master/lib/rocket_job/extensions/aasm.rb

By adding this capability to AASM we can remove the AASM patches from the Rocket Job gem.

anilmaurya:
Thank you @reidmorrison and @Tybot204 , I appreciate your efforts ♡

PR #424, opened 2016-12-21
Feature/contributing

whatisinternet:
Thanks for the great project! 

# What it does:
- Adds a contributing document to help level up new users

# Why it does it:
- The gem doesn't provide a clear and coherent message on how to contribute beyond the default template.

# Further discussion:
I still haven't been able to get all of the tests running green locally, which makes development hard. Although, I am a lot closer now than with the current documentation. Currently between 7 and 15 tests fail (randomly) on `master` locally.

I wanted to try and get the ball rolling on improvements to the contributing docs if possible since I'd like to help and I'm sure others would too. 😄  Please let me know if there is anything I've missed or anything that needs to be added. 

Thanks again!

PR #430, opened 2017-01-14
Resolve Ruby 1.9.3 failing on CI

whatisinternet:
# What it does:
- Resolves issues with Nokogiri and Ruby 1.9.3

# Why it does it:
- Ruby 1.9.3 issues as outlined [here](https://github.com/aasm/aasm/pull/424#issuecomment-272588809) on #424 

# Bigger questions:
Should this gem still support ruby 1.9.3? Version 5 feels like a really good point to EOL 1.9.3 support. Official support ended [Feb 2015](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/) [Secondary](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/). Nokogiri [ended support](https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md#170--2016-12-26) for everything up to and including ruby 2.0. SASS removed support in [3.5](http://blog.sass-lang.com/posts/560719).


anilmaurya:
@whatisinternet Thank you very much, I appreciate your help ♡.

I also think Version 5 is good point to end ruby 1.9.3 support. I will add this in version 5.0 TODO.

PR #432, opened 2017-01-20
Add section about "bang" methods to readme

iamvery:
First, apologies if this isn't entirely accurate. I stumbled upon this understanding and didn't see it laid out clearly in the readme.

Basically for some reason I had in my mind that the "bang" methods corresponded to error failures rather than persistence. The readme makes it obvious that errors are the default behavior, but it sort of glazes over the persistence thing.

Hope this helps! Thanks for your hard work on the library :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,15 @@ the transition succeeds :
   end
 ```
 
+### Bang Methods
+
+Each even method also comes in a "bang" flavor. The "bang" version, e.g. `job.run!`, automatically persists the record after a successful state transition. It is important to note that the "bang" in this case does not have anything to the event raising an exception or not.
+
+```ruby
+job.may_run? # => true
+job.run!     # => true, and the job has also been saved to the database
+```
+
 ### Callbacks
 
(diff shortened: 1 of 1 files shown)


anilmaurya:
Hi @iamvery , thanks for your effort.

We have a block explaining `!` methods

screen shot 2017-02-24 at 1 27 27 pm


May be we need to highlight it, to make it more readable.

iamvery:
@anilmaurya thanks for the reply. I do think your documentation would benefit from further highlighting of this behavior. We found it a little surprising initially that the "bang" indicated persistence rather than error (though we're not questioning that decision :). It would have been easier to find in the documentation if a search for "bang" on the page pin-pointed that precise point. I'll let you all figure out how to update the docs.

Thanks so much for your hard work on a great little library! ❤️

anilmaurya:
Thank you for reporting issue. 👍

Bang events is highlighted in https://github.com/aasm/aasm/commit/e1adfcd126a1f362c53e2440b71e8ddc7f5d7a86

PR #433, opened 2017-01-24
Search unscoped class when saving state without validation

larissa:
Records that are outside of `default_scope` on ActiveRecord don't transition when `skip_validation_on_save: true`. I fixed the issue by searching for the record with `unscoped.where` instead of just `.where`.

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -103,7 +103,7 @@ module AASM
       private
 
         def aasm_update_column(name, value)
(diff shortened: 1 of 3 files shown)


anilmaurya:
Hi @larissa , Thank you for reporting issue and sending patch.

Can you add specs for this case so that we don't break in future ?

larissa:
@anilmaurya sure! I'll add specs as soon as I get a chance.

larissa:
@anilmaurya I fixed specs and added a new one. Let me know if you need any changes.

anilmaurya:
@larissa Thank you for contributing 👍

PR #438, opened 2017-02-26
Add binding event

leanhdaovn:
Allow an event to be bound to another

```ruby
class Example
  include AASM

  aasm(:work) do
    state :processing

    event :start do
      transitions :from => :sleeping, :to => :processing
    end
    event :stop do
      transitions :from => :processing, :to => :sleeping
    end
  end

  aasm(:question) do
    state :answered, :initial => true
    state :asked

    event :ask, :binding_event => :start do
      transitions :from => :answered, :to => :asked
    end
    event :answer, :binding_event => :stop do
      transitions :from => :asked, :to => :answered
    end
  end
end

example = Example.new
example.aasm(:work).current_state #=> :sleeping
example.aasm(:question).current_state #=> :answered
example.ask
example.aasm(:work).current_state #=> :processing
example.aasm(:question).current_state #=> :asked
```

### DIFF
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -159,9 +159,8 @@ private
       yield if block_given?
     end
 
-    binding_state_machine_name = self.class.aasm(state_machine_name).state_machine.config.binding_state_machine
     binding_event = event.options[:binding_event]
-    if binding_state_machine_name && binding_event
+    if binding_event
       __send__("#{binding_event}#{'!' if persist}")
     end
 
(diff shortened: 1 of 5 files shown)


anilmaurya:
Hi @leanhdaovn ,

I think :binding_event makes sense when we have multiple AASM block in a class.

Thank you for contributing 👍

leanhdaovn:
Thanks @anilmaurya

alto:
@leanhdaovn Wow, very nice addition. Thanks a lot!

PR #439, opened 2017-03-07
Drop Support for Mongo Mapper

reidmorrison:
Mongo Mapper is now dead, and has been replaced by Mongoid.
Remove support for MongoMapper in newer versions of AASM.


reidmorrison:
@anilmaurya if you have any issues in the future related to AASM and its Mongoid persistence layer, feel free to reach out to me and will see what I can do. We use the AASM Mongoid persistence layer extensively in [Rocket Job](https://github.com/rocketjob/rocketjob) so have quite a bit of experience with it.

anilmaurya:
@reidmorrison Thank you for the helping hand.

I see lot of changes in gemfiles which I am not sure of, will go through the changes and let you know.

reidmorrison:
Yes, we also use Appraisal for all of our ruby gems that have multiple gemfiles.

With Appraisal it recommends placing all common gem requirements in `Gemfile` that way the `Appraisals` only have those specific gems that are needed. It greatly simplifies the `Appraisals` file.

I also removed the additional Rails 4.2 appraisal since it can do Mongoid v4 at the same time. The Mongoid 5 appraisal does not also need to re-test the other persistence layers already covered as part of Rails 4.2.

I also fixed the `.travis` file which already has newer Ruby versions, but it appears the exclusions list was not updated at the same time and was referring to older Ruby version that are no longer used.

Let me know, or email me directly if you have any questions about the Appraisal changes.

anilmaurya:
Appraisal is looking much cleaner now, Thank you for explaining 😊

PR #442, opened 2017-03-14
Make aasm thread-safe. 

reidmorrison:
Ever since AASM supported multiple named state machines in a single class, it has accessed and stored its state machines in class variables.

Accesses to the named state machines are not thread-safe causing exceptions when multiple threads attempt to access the same state machine at the same time during lazy initialization.
Using a Concurrent Map makes calls to state machines, and lazy initialization thread safe.

Unfortunately existing test frameworks do not support testing for concurrency race conditions. This patch is running in our production environment where we have not seen any exceptions due to this concurrency race condition since making this patch.

[concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby) is the concurrency library for ruby and is used for concurrency in many gems, including Rails.


reidmorrison:
Below is the exception that lead to the changes in this PR:

~~~
NoMethodError
vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/core/event.rb:115:in `block in _fire': undefined method `from' for :failed:Symbol (NoMethodError)
    from vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/core/event.rb:135:in `block in _fire'
    from org/jruby/RubyArray.java:1734:in `each'
    from vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/core/event.rb:127:in `_fire'
    from vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/core/event.rb:37:in `fire'
    from vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/aasm.rb:119:in `aasm_fire_event'
    from vendor/bundle/jruby/2.3.0/gems/aasm-4.11.1/lib/aasm/base.rb:126:in `block in fail!'
    from vendor/bundle/jruby/2.3.0/gems/rocketjob-3.0.0/lib/rocket_job/plugins/job/worker.rb:130:in `rocket_job_fail_on_exception!'
    from vendor/bundle/jruby/2.3.0/gems/rocketjob-3.0.0/lib/rocket_job/plugins/job/worker.rb:67:in `rocket_job_next_job'
    from vendor/bundle/jruby/2.3.0/gems/rocketjob-3.0.0/lib/rocket_job/worker.rb:108:in `process_available_jobs'
    from vendor/bundle/jruby/2.3.0/gems/rocketjob-3.0.0/lib/rocket_job/worker.rb:81:in `run'
    from vendor/bundle/jruby/2.3.0/gems/rocketjob-3.0.0/lib/rocket_job/worker.rb:48:in `block in initialize'
~~~

The AASM state machine for the above exception is available at: https://github.com/rocketjob/rocketjob/blob/master/lib/rocket_job/plugins/job/state_machine.rb#L22

anilmaurya:
Thank you @reidmorrison

alto:
@reidmorrison Perfect, thanks a lot!

PR #444, opened 2017-03-18
Use concurrent map to hold list of state machines

reidmorrison:
With the previous concurrency fix PR, I forgot this patch because we made it quite some time ago.

It ensures that accesses to the list of state machines is thread safe.


anilmaurya:
Thank you @reidmorrison for fixing these issues, I appreciate your contribution ♡

Issue #454, opened 2017-04-11
Incorrect guard checking when from is not provided


rafaelsales:
Version: 4.12.0

Expected this to work
```
event :work do
  transitions to: :foo, guard: :should_foo?
  transitions to: :bar, guard: :should_bar?
end
```
When `should_foo` returns `false`, the `should_bar?` is not even called, but when `from` is provided, the `should_bar?` is checked.
```
event :work do
  transitions from: :initial, to: :foo, guard: :should_foo?
  transitions from: :initial, to: :bar, guard: :should_bar?
end
```

adam-szczombrowski:
Version: 4.12.0

I've encountered the same problem. IMO the behaviour should be the same whether you specify the `:from` option or not. Will try to come up with a pr solving this issue.

adam-szczombrowski:
@anilmaurya can you confirm that this is a bug? Maybe it's desired behaviour and I don't understand something.

anilmaurya:
@adam-szczombrowski I agree that behaviour should be  same whether you specify the `:from `option or not . Looking forward to your PR.

Thank you for helping 😊

anilmaurya:
Fixed in #456 , thanks to @adam-szczombrowski

PR #456, opened 2017-04-27
Fix guard checking when from is not provided

adam-szczombrowski:
Fixing https://github.com/aasm/aasm/issues/454. Lack of `:from` option in transition resulted in incorrect guard checking. This pr solves the issue.


anilmaurya:
@adam-szczombrowski Thank you very much ♡

PR #460, opened 2017-05-19
Add header to deep link to lifecycle

agarcher:
I was referencing the AASM lifecycle in a PR I was writing and I wanted to deep link to the order of lifecycle events in the README, but there is no heading for it. This just adds a heading for linking. I also moved it down in the file a bit to keep the rest of the info on callbacks together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,37 +165,6 @@ class LogRunTime
 end
 ```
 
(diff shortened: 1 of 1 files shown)


anilmaurya:
Thank you @agarcher , it make sense to have header for lifecycle.

PR #462, opened 2017-05-27
Fix "Directly inheriting from ActiveRecord::Migration is not supported" error

pacuna:
I think this was introduced in the 5.1 release. I just added the current version to the migration.


anilmaurya:
@pacuna Thank you for sending patch 🎖

PR #482, opened 2017-07-11
Update Redis support

### LINKED ISSUE #480: ArgumentError when using AASM with Redis
I've got the following class:

```
class SomeClass
  include Redis::Objects
  include AASM

  value :aasm_state

  aasm do
    state :state_one, initial: true
    state :state_two

    event :forward do
      transitions from: :state_one, to: :state_two
    end

    event :reverse do
      transitions from: :state_two, to: :state_one
    end
  end

  def id
    1
  end

end
```

I can create a new instance of the class, and it has an initial state set correctly. When I try to trigger an event, I get the following an error...

```
> s = SomeClass.new
> s.forward!
ArgumentError: wrong number of arguments (given 2, expected 1)
from <my_path>/gems/aasm-4.12.0/lib/aasm/persistence/redis_persistence.rb:84:in `aasm_write_state'
```

Am I missing something here?

focusshifter:
Fix #480.


anilmaurya:
Thank you @focusshifter 🏆

PR #483, opened 2017-07-11
Enable Dynamoid for Rails 5 specs

focusshifter:
Dynamoid received Rails 5 support according to https://github.com/Dynamoid/Dynamoid/issues/43#issuecomment-310518341, so we may reenable it in Appraisals.


anilmaurya:
Thank you @focusshifter 👍

Issue #449, opened 2017-04-04
may_*? method `wrong number of arguments (given 1, expected 0)`


kntmrkm:
- AASM 4.12.0
- Rails 5.0.1

`may_contract?` method will be got below error.

```
ActionView::Template::Error (wrong number of arguments (given 1, expected 0)):
```

The error occur with below code.

```
    event :contract do
      transitions from: [:authorized], to: :contracted, guard: :can_contract?
    end

  ・・・・

  def can_contract?
    self.authorizer_id.present?
  end

```

But when rewite with below code. The error disappears.

```
    event :contract do
      transitions from: [:authorized], to: :contracted do
        guard do
          self.can_contract?
        end
      end
    end
```

klaustopher:
I got the same issues, when executing my test suite (mostly `it { is_expected.not_to allow_transition_to(:some_state) }`).

Everything works fine with 4.11.x

hirox:
Hello. I have no idea to fix it, but I found the cause of the error.

With this commit, https://github.com/aasm/aasm/commit/c3b6454e8e9ca499e57aac4c74ec18bf9fa028b4#diff-6323dc24ff3a8a3e52b3b26d63e926cdR126 (#404, #406), `args.unshift(nil) if args.blank?` was added and `args` will be `[nil]` although there is no arguments.
This change causes validation error because `can_contract?` expects 0 arguments.

klaustopher:
It's very strange, that this issue has been open for 22 days and nobody who's involved in the project has even acknowledged that a bug was introduced that breaks behavior that is documented in the readme.

anilmaurya:
Hi @klaustopher , Thank you for keeping eye on this issue.

I am running of time for last few weeks. Any help is much appreciated.

A failing test will help in solving issues quicker.

anilmaurya:
Hi @kntmrkm @klaustopher ,

I added a spec to test this scenario for ActiveRecord.

https://github.com/aasm/aasm/blob/master/spec/unit/guard_spec.rb#L86

Specs are passing without failure, may be I missed something.

Let me know what do you think

hirox:
Hello @anilmaurya ,
I found it occurs when rspec-mocks is used.
rspec-mocks destroys arity (from 0 to -1) and `record.__send__(code, *args)` (args == [nil]) is called here ( https://github.com/aasm/aasm/blob/master/lib/aasm/core/transition.rb#L72 )

Here is a backtrace.
```
[...]

Issue #441, opened 2017-03-13
ArgumentError: wrong number of arguments when first argument is nil


werleo:
When first argument to event is nil it is removed from argument list.
In consequence we get:

```
ArgumentError: wrong number of arguments (given 1, expected 2)
```

Test scenario:

https://github.com/werleo/aasm/commit/354e05f27f49ee15eaca1d8649b369b971d3e114

lichtamberg:
I think i found the problem, working on a patch at the moment

lichtamberg:
It looks like its this method definition which removes the first argument, because to_state is nil:

```ruby
#aasm/lib/aasm/core/event.rb @ line 44 AASM::Core::Event#may_fire?


    43: def may_fire?(obj, to_state=nil, *args)
 => 44:   binding.pry if args.size == 1
    45:     puts "#{caller[0]}\t#{args.to_s}" # debug
    46:   _fire(obj, {:test_only => true}, to_state, *args) # true indicates test firing
    47: end

[1] pry(#<AASM::Core::Event>)> to_state
=> nil
[2] pry(#<AASM::Core::Event>)> args
=> ["yuppi"]
```

Don't know how to refactor this though...

anilmaurya:
@werleo Can you run your specs on master branch and let me know if this issue is resolved ?

werleo:
@anilmaurya test is passing on master branch.
Thank you for fixing it.

Test is in PR https://github.com/aasm/aasm/pull/485/files
I made it by mistake so you may close it.

anilmaurya:
I would like to merge #485 , can you rename `wrong_arguments_list` to  `guard_arguments_check` ?

werleo:
Renamed.

Issue #416, opened 2016-11-16
Getting "wrong number of arguments (given 2, expected 1)" error, when using active record boolean check method as guard


PragmaticEd:
Getting this error:

```
wrong number of arguments (given 2, expected 1)
```

When executing an event, when using active record boolean check method as guard.
For example, i have `boolean` field in database, called `vip`.
Active record provides method `vip?`, that works fine everywhere, except, if you pass it as guard method:

```
# code that doesn't work
aasm :state do
  event(:pay)  {
    transitions(from: :finished, 
                     to: :paid,
                     after: [:handle_payment],
                     guards: [:vip?]
  )}
end

private

def handle_payment(argument)
end

```

Can be solved by making new private method, that does the check, like so:

```
# code that works
aasm :state do
  event(:pay)  {
    transitions(from: :finished, 
                     to: :paid,
                     after: [:handle_payment],
                     guards: [:is_vip?]
  )}
end

private

def handle_payment(argument)
end

def is_vip?
  self.vip?
end
```

But, i'm guessing, this is a bug, right?

I have to note, this error only occurres, if there is an `after` method that accepts an attribute:

```
@order.pay!(nil, "some value")
```

If the method `handle_payment` wouldn't accept arguments, then everything works fine.

ruby: `2.3.0p0`
rails: `5.0.0`
os:
```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
```

mvz:
I can confirm this with aasm 4.12.0 (I haven't checked the bit about having an `after` method with arguments). Using the attribute method without the `?` directly is also a good work-around (in this case, that would be `guard [:vip]`).

project-eutopia:
I ran into the same problem trying to upgrade to version 4.12.  I was able to isolate this as a problem with how `aasm` seems to be handling guard methods that have variable arguments.  The boolean check methods like `vip?` have arity -1 (variable) but when `aasm` uses it as a guard it passes along a single argument, `nil`.  Here is some sample code that isolates the problem of a variable argument guard method receiving an extraneous `nil` argument:

```ruby
require "aasm"

class Foo
  include AASM

  aasm do
    state :null, initial: true
    state :alpha
    state :beta

    event :make_alpha do
      transitions from: :null, to: :alpha, if: :check?
    end

    event :make_beta do
      transitions from: :null, to: :beta, if: :good?
    end
  end

  def check?
    true
[...]

PR #484, opened 2017-07-26
Use AASM::NO_VALUE as `to_state` argument default value

teohm:
Based on discussion in https://github.com/aasm/aasm/issues/449#issuecomment-318022061

This PR attempts to fix the root cause of a few issues (#449, #416, #441) with this similar error:

```
Wrong number of arguments (given 1, expected 0)
```

This error happens because AASM incorrectly calls guard method with
one extra argument in certain scenarios. For example:

```ruby
def can_do_something?  # guard method with no argument
  ...
end

# But AASM incorrectly triggers `can_do_something?(nil)`
```

This happens because in `AASM::Core::Event#_fire(obj, options={}, to_state=nil, *args)`,
the `to_state` argument default nil value is causing ambiguity, i.e. there is no way for 
AASM to tell whether a caller provides a) one argument with nil value, or b) no argument.

This PR removes this ambiguity by using a symbol as `to_state` default value.

### DIFF
--- a/spec/unit/complex_multiple_example_spec.rb
+++ b/spec/unit/complex_multiple_example_spec.rb
@@ -39,18 +39,18 @@ describe 'when being unsuspended' do
   it 'should be able to wait into waiting if polite' do
     auth.left_suspend!
     expect(auth.may_left_wait?(:waiting, :please)).to be true
-    auth.left_wait!(nil, :please)
+    auth.left_wait!(:please)
 
     auth.right_suspend!
     expect(auth.may_right_wait?(:waiting)).to be false
-    auth.right_wait!(nil, :please)
+    auth.right_wait!(:please)
   end
 
   it 'should not be able to be unsuspended into active if not polite' do
     auth.left_suspend!
     expect(auth.may_left_wait?(:waiting)).not_to be true
     expect(auth.may_left_wait?(:waiting, :rude)).not_to be true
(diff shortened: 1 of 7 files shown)


spec/unit/complex_multiple_example_spec.rb
```diff
@@ -39,18 +39,18 @@
   it 'should be able to wait into waiting if polite' do
     auth.left_suspend!
     expect(auth.may_left_wait?(:waiting, :please)).to be true
-    auth.left_wait!(nil, :please)
```
teohm: @anilmaurya I noticed `nil` is explicitly included in argument when triggering event in the specs. Is this a common in actual usage? This is different from the usage examples written in https://github.com/aasm/aasm/wiki/State-transitions-with-guard-clauses-that-require-arguments

anilmaurya: @teohm Including `nil ` in argument list seems weird to me.

anilmaurya:
Thank you @teohm for contributing , I appreciate your help 🎖

teohm:
@anilmaurya By any chance to release this fix some time this month August? 🙏

anilmaurya:
Hi @teohm ,

I am planning to release new version with this fix soon.

teohm:
That's great! thanks @anilmaurya !! 🎉

anilmaurya:
Hi @teohm , just release `4.12.2` with this fix.

teohm:
Thanks @anilmaurya !!

PR #465, opened 2017-05-30
Add use transaction flag (README pending)

lichtamberg:
I've added a flag wrapping of transactions with ActiveRecord.. See also #431 
Please let me know if this is fine and I will add more README stuff.

PR #493, opened 2017-08-08
Add use transaction flag

### LINKED ISSUE #431: How to skip transactions with ActiveRecord persistence
Our application performs a "create" operation which cannot be rolled back without side effects. Specifically, we call an external API to create an object in another system, then add the corresponding record in our database. If an exception is raised, the record is rolled back, but the object is not deleted.

Right now we are hacking around this by performing the operation in a `before_transaction` block, but it will be difficult to maintain going forward. Is there a way to skip transactions with ActiveRecord persistence? If not, could you add one?

mbramson:
Adds a flag that allows state machines to not use ActiveRecord transactions at all.

This PR is basically #465, so almost all credit goes to @lichtamberg. I simply solved the merge conflict and added this feature to the README.

Closes #431

### DIFF
--- a/README.md
+++ b/README.md
@@ -867,6 +867,24 @@ end
 
 which then leads to `transaction(:requires_new => false)`, the Rails default.
 
+Additionally, if you do not want any of your active record actions to be
+wrapped in a transaction, you can specify the `use_transactions` flag. This can
+be useful if you want want to persist things to the database that happen as a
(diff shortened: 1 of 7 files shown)


mbramson:
@lichtamberg and @anilmaurya, when you get a chance, take a look!

anilmaurya:
Thank you @lichtamberg and @mbramson ♡

PR #499, opened 2017-09-07
Fix job.fire to job.aasm.fire in README

pgericson:
The Pull request that introduced this #494 wrote the wrong instance on what the method `state_event_instance.fire` could be triggered.


anilmaurya:
Your are right @pgericson , Thank you

PR #517, opened 2017-12-19
Describe the correct option to use a custom AASM base on a column

mnd-dsgn:
The docs and the code example differ. The code example has the correct implementation.


anilmaurya:
@mnd-dsgn Thank you for fixing 👍

PR #518, opened 2017-12-20
Proposal: set state machine name to default

cathykc:
We use `aasm_fire_event` in a bunch of places in our codebase and it doesn't make sense to keep passing in `:default` for the `state_machine_name` arg. Proposing that we default the argument to `:default`.

If this is seems like a positive change I'll spend more time and write tests.


cathykc:
Never mind - looks like we should replace direct uses of `aasm_fire_event` with `fire` and `fire!`

Issue #526, opened 2018-02-01
Should `error` callback return `false` or the actual return value?


zinosama:
**Context:**

```ruby
aasm column: :state, whiny_transitions: false do
  state :new
  state :running

  event :run, before: :prepare do
    error :handle_error
    transitions from: :new, to: :running
  end
end

def prepare
  raise "leg hurts"
end

def handle_error(e)
  log(e)
  false
end
```

**Current Behavior:** 
```ruby
obj.state # => "new"
obj.run!   # => true
obj.state # => "new"
```

**My Expected Behavior:**
```ruby
obj.state # => "new"
obj.run!   # => false
obj.state # => "new"
```

**The Reason I Expect That Behavior:**
1. In the readme, it says 

> If whiny_persistence flag is set to false, methods with a bang return true if the state transition is successful or false if an error occurs.

which sounds to me like if an error occurs, it should either return the return value of my callback or simply `false`.

2. Since `run!` causes a mutation, the return value should signal whether the action is performed correctly. Conceptually, it's confusing to have a return value of true when the action actually failed.

3. Implementation-wise, it makes error detection difficult since in order to determine whether the mutation went correctly, I'd have to collect and compare the before and after values. Or, I'd have to re-raise the exception and catch it everywhere I call `obj.run!`.

I've read the code, and I understand the mechanism of the current behavior. However, could someone please explain the reason for this design? Or whether a PR will accepted to change this behavior?

Thanks a lot in advance.

anilmaurya:
Hi @zinosama ,

I completely agree that if an error occurs then state transition should  return false.

I am happy to merge PR for this .

Thank you for contributing and reaching out .

PR #527, opened 2018-02-03
Always return false when error callback is invoked

zinosama:
Fix https://github.com/aasm/aasm/issues/526


zinosama:
@anilmaurya Let me know what you think 😄 Thanks!

anilmaurya:
@zinosama Thank you for contributing. 👍

PR #530, opened 2018-02-13
Run Travis tests with Ruby 2.4.2

alto:


nicolasleger:
You can also add Ruby 2.5.
However Rails 4.2 seems to have some problems with Ruby >= 2.4.

PR #537, opened 2018-03-02
chore(invokers): Refactor event invokers (Initial ver)

pandomic:
* Separate invoking logic to `Invoker` class
* Introduce class-based invokers
* Introduce sub-invoking classes for each invoke-type: Proc, String, Symbol
* Add capability with `failures` and different default return values

## Breaking changes
* To be convenient, state proc-callbacks are now executed in a `record`'s context as well.


pandomic:
This is a draft which I put in aasm repo by mistake. It's is not ready to be there.

PR #539, opened 2018-03-02
Circumvent "instance variable not defined" warnings

### LINKED ISSUE #583: Localizer (l18n) only works if the class responds to :model_name
`AASM::Localizer` will only translate classes that respond to `:model_name`

IMNSHO it should work on any class that can use AASM.  

I realize that does introduce an issue as to how the localization file should be structured. I'd suggest that AASM could adopt the same structure as is used by Rails, but just remove the `active_record' level.

Ex:

instead of this (taken from `spec/en.yml` and used in localizer_spec.rb): 
```
en:
  activerecord:
    events:
      localizer_test_model:
        close: "Let's close it!"

  attributes:
    localizer_test_model:
      aasm_state/opened: "It's open now!"
```

it would be **this**:
```
en:
  events:
    localizer_test_model:
      close: "Let's close it!"

  attributes:
    localizer_test_model:
      aasm_state/opened: "It's open now!"
```

The `en.activerecord` level is removed and everything below it is shifted 'up'.

This has the advantage of being familiar to those who have worked with Rails I18n, and also providing a well-defined specification to work with.

If AASM is not going to support localizing (l18n/i18n) of classes that don't respond to `:model_name` then that should be documented in the README.

pcantrell:
Fixes #583.

### DIFF
--- a/lib/aasm/persistence/plain_persistence.rb
+++ b/lib/aasm/persistence/plain_persistence.rb
@@ -5,7 +5,8 @@ module AASM
       # may be overwritten by persistence mixins
       def aasm_read_state(name=:default)
         # all the following lines behave like @current_state ||= aasm(name).enter_initial_state
-        current = aasm(name).instance_variable_get("@current_state_#{name}")
+        current = aasm(name).instance_variable_defined?("@current_state_#{name}") &&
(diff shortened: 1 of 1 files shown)


anilmaurya:
@pcantrell Thank you for contributing ♡

Issue #503, opened 2017-10-03
Class style callbacks do not work for events


Averethel:
I can declare a class to handle callback for transitions:

```ruby
transition from: :waiting, to :declined, after: Nofitier
```

but when I want to do the same for events (for example to have the correct state in the `Notifier`)
```ruby
event :decline, after: Notifier
  transition from: :waiting, to :declined
end
```
it does not work. I would expect this interface to be consistent between events and transitions.

TJC:
I spent a while trying to figure out what was wrong with my code, and came to the same discovery.

It's actually worse -- the class-style callbacks don't work on all events for transitions.

```
# This calls MyLogEvent:
transition from: :waiting, :to :declined, :after MyLogEvent

# This does not:
transition from: :waiting, :to :declined, :success MyLogEvent
```

pandomic:
There are no class-handlers in event invokers for the most of state/transition/event callbacks :(

PR #541, opened 2018-03-18
chore(invokers): Refactor callback invokers, add class-callbacks support

pandomic:
## Motivation
While working on a big project we faced with a problem - our models started to grow, we needed to make a huge processing on state/transition/event callbacks,  and keeping that logic in procs or in a model methods was not a good solution for us. Currently we are using a monkey-patched version of invoking logic to add a class-callbacks support. But I guess there also might be people who faced the same problem and need this feature supported officially.

While diving into the AASM code we found that invoking logic is mostly duplicated, so we decided to extract duplicated parts and add a class-callbacks support.

Related issue: https://github.com/aasm/aasm/issues/503

## A few words about PR
PR contains base invoking class and concrete invokers for each subject type (Proc, Literal, Class). Base invoker decides which concrete invoker to use and also takes some additional configuration to support states/events/transitions (as they may have some differences in default return values, options and etc.)

## Breaking changes
* To be convenient, state proc-callbacks are now executed in a record's context as well.


anilmaurya:
Hi @pandomic .

Thank you for raising this PR. I will go through changes over the coming weekend.

anilmaurya:
@alto Need your review on this. Can you have a look once ?

alto:
@anilmaurya Sure.

lib/aasm/core/state.rb
```diff
@@ -13,7 +15,13 @@ def initialize(name, klass, state_machine, options={})
     def initialize_copy(orig)
       super
       @options = {}
-      orig.options.each_pair { |name, setting| @options[name] = setting.is_a?(Hash) || setting.is_a?(Array) ? setting.dup : setting }
+      orig.options.each_pair do |name, setting|
```
alto: Reformatted to make easier to read 👍

pandomic:
Thank you guys, will try to make some updates during the weekend

pandomic:
Hey @alto, please take a look one more time when you will have some time ;)

alto:
@pandomic Looks good now. Cheers!

anilmaurya:
@pandomic Can you please resolve conflicts ?

anilmaurya:
Hi @pandomic ,

Thank you for your contribution.
I have merged your PR and released AASM `5.0.0` .
🏅🏆

pandomic:
Hi @anilmaurya, thank you a lot! Sorry, had no time to resolve conflicts, hope there was not too many

tisba:
Nice change 👍 

Just a quick question: `LogRunTime`  is now defined twice in the readme (https://github.com/aasm/aasm#callbacks). Is this an oversight or is this to show a special behaviour? Sorry if this might be a stupid questions, I'm trying to wrap my head around this :)

pandomic:
oh, yeah, looks like the first one is not needed anymore, as the second one explains behavior a bit more in detail

anilmaurya:
@tisba Thanks for pointing it out. 👍

Issue #555, opened 2018-04-30
Get the guards for a non-permitted event


cdesch:
How do I retrieve the guards for the non-permitted events for a state? For example, let's say I have the Job class from the ReadMe
```

class Job
  include AASM

  aasm do
    state :sleeping, :initial => true
    state :running, :cleaning

    event :run do
      transitions :from => :sleeping, :to => :running
    end

    event :clean do
      transitions :from => :running, :to => :cleaning, :guard => :cleaning_needed?
    end

    event :sleep do
      transitions :from => [:running, :cleaning], :to => :sleeping
    end
  end

  def cleaning_needed?
    false
  end
end
```

We will create a new job and set it to `running`

```
job = Job.new
job.run
```

We can get the non permitted events with 
```
job.aasm.events(:permitted => false).map(&:name)
#=> [:clean]
```

How can I get the guard `:cleaning_needed?` from the event as the condition(s) that need to be satisfied in order to allow the event to trigger?

`@guards` from `ASM::Core::Event` is not directly accessible to retrieve the current guards required for the event.

timlawrenz:
@cdesch Can this Issue be closed?

PR #556, opened 2018-05-01
Adding failed_callbacks example

cdesch:
Adding `failed_callbacks` example to the Inspection Section
Resolves issue #555 in terms of AASM knowledge :-)


cdesch:
This is not exactly what it does as the function lists all the failed callbacks.... I think I need to take a break for the day. I have some code I'll submit later for a PR.

PR #559, opened 2018-05-18
Fix and update README.md

ikataitsev:
This PR fixes an example in https://github.com/aasm/aasm/blob/master/README.md#handling-naming-conflicts-between-multiple-state-machines.


anilmaurya:
Thanks @ikataitsev

PR #560, opened 2018-05-24
Enable AASM scopes to be defined on abstract classes

RickCSong:
Currently, if aasm is called on an abstract class, the scopes are not
properly defined because it infers the table_name at time of creation.
This change pushes the inference of table_name to when the scope is
called which enables scopes to be dynamically defined in descendant
models.


anilmaurya:
Hi @RickCSong .
Thank you for fixing this issue. Can you help with fixing travis build ?

RickCSong:
Fixed!  Thank you.  Just found out that scopes on abstract classes didn't work until ActiveRecord 5. Added a comment.

RickCSong:
Hey there @anilmaurya !  Any chance you may be able to take another look at this?

anilmaurya:
Thanks @RickCSong 👍

PR #564, opened 2018-05-28
Set current event on fire (Fixes 551)

### LINKED ISSUE #551: aasm.current_event incorrectly yields nil when calling aasm.fire!(:event)
Given the following code excerpt:

```
aasm column :state do

  state :unkown, initial: true
  state :matched

  event :match do
    transitions from: :unknown, to: matched, after: :matched_event
  end

  def matched_event
    puts aasm.current_event  # yields nil if triggered with aasm.fire!(:match)
  end

end

aasm.match!                # => Correct, aasm.current_event yields 'match!'
aasm.fire!(:match)         # => Incorrect, aasm.current_event yields nil
```

Would be nice to get this fixed, as my controllers are beginning to look a lot like hackery (i have 30+ transitions) and have had to resort to this less than elegant code...

```
def update
  ...
  if model.respond_to?("may_#{params[:event]}?') && model.send("may_#{params[:event]}?")
    model.send("#{params[:event]}!")
  end
  ...
end
```

TikiTDO:
Updates the `fire` and `fire!` methods to set the current event. Fixes #551

### DIFF
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -104,10 +104,14 @@ module AASM
     end
 
     def fire(event_name, *args, &block)
+      aasm_name = @name.to_sym
+      @instance.aasm(aasm_name).current_event = :"#{event_name}"
       @instance.send(:aasm_fire_event, @name, event_name, {persist: false}, *args, &block)
     end
 
     def fire!(event_name, *args, &block)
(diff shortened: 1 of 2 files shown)


iMacTia:
@TikiTDO thanks for this fix! We just encountered this same issue.
Hopefully @anilmaurya can merge this soon 👍

PR #580, opened 2018-10-03
Add basic neo4j persistence

lshimokawa:


anilmaurya:
Hi @lshimokawa , thanks for your contribution.

Can you please add specs so that we don't break it in future

PR #594, opened 2019-01-16
update Readme: add headings, improve formatting

BKSpurgeon:
### What is this change?

* It is a simply change in the formatting and presentation: bullet points have been added, and an extra heading has been added.

### Why this change?

* to make it clearer to the user: I was searching for a bug when I realised that a simple 'require' statement was omitted in code. This change request makes it explicitly clear so that it basically cannot be overlooked by future users. I hope to spare them some pain.

I hope this helps. chrs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1178,7 +1178,15 @@ the 'instance method symbol / string' way whenever possible when defining guardi
 
 #### RSpec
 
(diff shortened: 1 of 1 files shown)


anilmaurya:
Thanks for contribution 👍

Issue #584, opened 2018-10-30
Events accept invalid "to" states


sypher7:
### Description
Attempting to pass a `to_state` to an event that is not defined has unexpected results (it is interpreted as an argument, and a `nil` `to_state` is used).

### Steps to reproduce
1. Fire an event with an invalid `to_state` as the first parameter:
```
some_state_machine.run!(:foo)
```

### Expected Result
`AASM::InvalidTransition` should be raised.

### Actual Result
The first defined transition for the `from_state` is used, similar to how passing `nil` works.

### Digging Deeper
The source of this appears to be the code here:
https://github.com/aasm/aasm/blob/d23e75ef46c3a3f1a6bd66dba780e1fac506e44b/lib/aasm/core/event.rb#L131-L139

If the `to_state` argument to the `_fire` method is not `::AASM::NO_VALUE` or included in the event transition's `to` states, AASM assumes that this is simply part of the argument list.

### Solutions
This could probably be refactored a bit. Perhaps a check to see if the second argument is an options `Hash`, and allow for a `to_state` option that could be explicitly set, with the rest of the options passed through.

In lieu of any sort of major change here, the documentation could probably be updated:

> You can also pass in `nil` as the desired end state, and AASM will try to transition to the first end state defined for that event.

A caveat about invalid states could be added to expand on that.

I'd be happy to open a PR after some discussion on what should be done (if anything) about this.

Thanks!

anilmaurya:
@sypher7 Thank you for reporting .
@skolsuper updated docs, therefore closing this issue.

PR #596, opened 2019-01-23
Update docs to more accurately reflect behaviour

skolsuper:
Usage of the word "must" in the current documentation is not accurate, if the first argument to a transition is not a valid end state, `to_state` will be set to `nil` and the arg will be unshifted to the args that get passed to callbacks: 

https://github.com/aasm/aasm/blob/cd25a1eeaeaec49238ec19b6cdd2b28d23c56f71/lib/aasm/core/event.rb#L135-L139

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Also, you can pass parameters to events:
 
 In this case the `set_process` would be called with `:defragmentation` argument.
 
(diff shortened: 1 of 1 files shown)


skolsuper:
Also relates to #584

anilmaurya:
@skolsuper Thank you correcting docs 👍

PR #597, opened 2019-01-25
Clarify `on_event` can accept *multiple* events

jackscotti:
It was clear it could have accept arguments as the previous examples already did cover the case. This chagne clarifies what the examples is covering.


anilmaurya:
Thanks @jackscotti

PR #621, opened 2019-04-23
Allow state defined with any number of :state args and optional options hash

tjgfernandes:
Allow state defined with any number of :state args and optional options Hash as last argument.

aasm example: 
`
state :state1, :state2, :state3, after_enter: :run_update
`

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -97,7 +97,8 @@ module AASM
     # [1] options (or nil)
     # or
     # [0] state
-    # [1..] state
+    # [1..N-2] state
+    # [..N-1] options hash (or nil)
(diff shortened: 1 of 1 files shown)


anilmaurya:
@tjgfernandes Can you explain use case for this change ?

tjgfernandes:
Hi @anilmaurya 

Currently if you need to run same method on 2 different states you do somthing like this:
```
state :running, after_enter: :run_update
state :walking, after_enter: run_update
```

with the proposed change you can do:
`state :running, :walking, after_enter: run_update`

anilmaurya:
Nice.

Can you add specs and update Readme as well ?
Thank you for contributing 👍

PR #626, opened 2019-05-09
Minitest - ensure assertion does not raise if transition is invalid (return false instead)

marceloeloelo:
Given a state machine:

```ruby
class Cleaner
  include AASM

  aasm do
    state :idle, initial: true
    state :cleaning

    event :clean do
      transitions from: :idle, to: :cleaning, guard: :cleaning_needed?
    end
  end

  def cleaning_needed?
    false
  end
end
```

A test that does:

```ruby
cleaner = Cleaner.new
refute_transitions_from cleaner, :idle, to: :cleaning, on_event: :clean
```
will raise a `AASM::InvalidTransition` exception. 

I believe a nicer behavior would be for the test to pass, instead.


anilmaurya:
Hi @marceloeloelo ,

Thank you for raising this PR. I agree test should pass instead of raising `AASM::InvalidTransition` exception.

Can you please add specs for this in `minitest_matcher_test` ?

Thanks.

Issue #508, opened 2017-10-27
Undefined method "blank?" introduced as of 4.12.1, persists in 4.12.3


fdr:
Hello,

I have a project that works perfectly with `4.12.0`, but when I upgrade aasm to `4.12.1`, `4.12.2`, or `4.12.3`, I receive stack traces complaining about `blank?` being undefined.

This is part of `activesupport` that is not `require`d by default. I can work around it by changing my gemfile to have this:

```ruby
gem 'activesupport', require: 'active_support/core_ext/object/blank'
```

But it does seem sub-ideal that AASM will crash out the gate otherwise.

anilmaurya:
Hi @fdr ,

Can you provide full error trace and steps to reproduce this error ?

If you can provide a failing specs then it will help us resolve it sooner.

Thanks, Anil.

fdr:
Sure. It is with any use of the `statename?` predicates:

```
NoMethodError: undefined method `blank?' for "running":String
```

```ruby
["/home/fdr/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/aasm-4.12.3/lib/aasm/persistence/base.rb:37:in `aasm_read_state'",
 "/home/fdr/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/aasm-4.12.3/lib/aasm/instance_base.rb:12:in `current_state'",
 "/home/fdr/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/aasm-4.12.3/lib/aasm/base.rb:103:in `block (2 levels) in state'",
 "(pry):1:in `<main>'",
]
```

will:
@fdr did you ever find a solution for this?

will:
I ran into this myself just now, and solved it by requiring the blank active support extension. The project I'm on does not use all of active support. It is not a rails project.

anilmaurya:
@will @fdr Would you like to raise PR to resolve this issue ?

tristandruyen:
FYI I made a PR which fixes this by using:
```
!thing || thing.empty?
```

instead of `thing.blank?`

anilmaurya:
Released new version `5.0.5` which includes fix for this issue. Thanks to @tristandruyen .

PR #627, opened 2019-05-16
Get rid of unneded activesupport base ext dep

tristandruyen:
This PR get's rid of the undocumented & unrequired activesupport dependency discovered in #508 by using builtin ruby stuff instead of activesupports `.blank?`


tristandruyen:
One could also solve this issue by explicitly adding activesupport to the gemsec and requiring blank...
I don't think this is needed as it's only used in a few places and was easy to replace.

anilmaurya:
Hi @tristandruyen 
Thank you for sending PR. I am wondering if we can write a spec to prevent using `blank?` in future.
what do you think ?

tristandruyen:
Sorry for the force-push fuckup...

I just added an appraisal which explicitly does not install rails.
You can see it fail here:(https://travis-ci.org/tristandruyen/aasm/builds/533703927)

It should fail when any core extensions added by rails is used, not just `blank?`, so this seems the safest way to test this IMO.

anilmaurya:
Great 🏆

Thank you, I Appreciate your contribution. ♡
I will merge this PR after travis build passes.

tristandruyen:
Awesome! Thanks for the quick merge 👍 
Could you please publish a new version when you find the time.
Thx alot 🥇

anilmaurya:
Yes, will release new version soon.

anilmaurya:
Released new version `5.0.5`

PR #638, opened 2019-07-14
Fix current event through fire

### LINKED ISSUE #551: aasm.current_event incorrectly yields nil when calling aasm.fire!(:event)

DoubleJarvis:
Should fix #551. It already has a PR that should fix it, but it has duplicate logic, and its just lying there for a year, unmerged, and the issue is still present.

### DIFF
--- a/spec/unit/event_spec.rb
+++ b/spec/unit/event_spec.rb
@@ -296,13 +296,15 @@ describe 'current event' do
   end
 
   describe "when calling events with fire/fire!" do
-    it "fire should populate aasm.current_event" do
+    it "fire should populate aasm.current_event and transition (sleeping to showering)" do
       pe.aasm.fire(:wakeup)
       expect(pe.aasm.current_event).to eq :wakeup
+      expect(pe.aasm.current_state).to eq :showering
     end
-    it "fire! should populate aasm.current_event" do
+    it "fire! should populate aasm.current_event and transition (sleeping to showering)" do
       pe.aasm.fire!(:wakeup)
       expect(pe.aasm.current_event).to eq :wakeup!
+      expect(pe.aasm.current_state).to eq :showering
     end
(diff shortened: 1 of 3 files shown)


spec/unit/event_spec.rb
```diff
@@ -294,6 +294,17 @@
     pe.wakeup!
     expect(pe.aasm.current_event).to eql :wakeup!
   end
+
+  describe "when calling events with fire/fire!" do
+    it "fire should populate aasm.current_event" do
+      pe.aasm.fire(:wakeup)
+      expect(pe.aasm.current_event).to eq :wakeup
```
anilmaurya: Can you add assertion for transition as well?

DoubleJarvis:
Are the tests in cf75bdb satisfactory, or you mean something else by transition? @anilmaurya

anilmaurya:
@DoubleJarvis Tests looks good 👍. Thank you for contributing.

Tests are failing on travis for some other reason. First I will fix travis build then I will merge your PR.

PR #639, opened 2019-07-16
Raise proper error when fire an unknown event

### LINKED ISSUE #618: fire! and fire with unknown event, crash with NoMethodError
When I run `fire` or `fire!` with an unknown event (i.e. `Model.aasm.fire!(:whatever)` ) it raises a `NoMethodError (undefined method fire_callbacks' for nil:NilClass)`, seems that it tries to fire before callbacks even not knowing that the event exists, which is not the expected behaviour.

For me, `fire` and `fire!` with an unknown event should not fire callbacks and return an exception like "Unknown event", to be consistent.

lesin:
Should fix #618 

if an unknown event was fired:
- raise `AASM::UndefinedState` 
- don't run callbacks

### DIFF
--- a/lib/aasm/aasm.rb
+++ b/lib/aasm/aasm.rb
@@ -96,6 +96,7 @@ private
 
   def aasm_fire_event(state_machine_name, event_name, options, *args, &block)
     event = self.class.aasm(state_machine_name).state_machine.events[event_name]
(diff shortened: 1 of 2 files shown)


anilmaurya:
@lesin Thank you for contributing
looks like there is conflicts and test cases are failing, can you please fix it ?

Issue #640, opened 2019-07-16
Running tests with Docker doesn't work


lesin:
Several issues occured while building an image.
Ex. issue with permissions and with locales:

<img width="1132" alt="aasm1" src="https://user-images.githubusercontent.com/1833428/61294320-da120f00-a7d5-11e9-9c63-c2fff658b267.png">
<img width="1134" alt="aasm2" src="https://user-images.githubusercontent.com/1833428/61294328-de3e2c80-a7d5-11e9-811e-2caac45475ee.png">

anilmaurya:
Fixed in #641

PR #641, opened 2019-07-16
Fix docker setup

lesin:
Fix for #640

I was trying to make the `msati/docker-rvm` image work but after fixing the issue with permissions and locales, I still wasn't able to run tests in the created container. 

Revert the usage of a previously used image `ruby:2.3.4-slim`.


anilmaurya:
@lesin Thank you for contributing.
I also tried to make `msati/docker-rvm` image work but it didn't worked out.
Your solution is working for me as well. My only concern is we won't be able run test for multiple ruby version on single container.
Running test for one ruby version is good than build error, therefore merging your PR.
Will investigate later on to support multiple ruby version.

PR #649, opened 2019-09-24
Update README.md

BKSpurgeon:
add further information re: column names when you have multiple state machines. I'm no expert, but please do confirm that we need to explicitly write in the column names when we have more than one state machine operating in a single model? Pls confirm.


anilmaurya:
Thank you @BKSpurgeon

PR #653, opened 2019-11-06
handle AASM::InvalidTransition in RSpec `transition_from` matcher

ryanwood:
When using the `transition_from` RSpec matcher, it returns false when the "from" state is valid but raises an exception when the "from" state is not valid. This commit aligns the two paths so the matcher will return false in both scenarios.


anilmaurya:
Thanks for contributing @ryanwood

PR #665, opened 2020-02-10
Adds `after_commit` for states and implements `after_all_commits`

RickCSong:
I added `after_commit` to make it easier to add callbacks that require a commit to happen after various state changes happen (i.e. enqueueing a job and preventing the race condition there).

Furthermore, I added `after_all_commits` since it appears the required code is already there (see lib/aasm/persistence/orm.rb)

### DIFF
--- a/README.md
+++ b/README.md
@@ -936,13 +936,15 @@ Since version *3.0.13* AASM supports ActiveRecord transactions. So whenever a tr
 callback or the state update fails, all changes to any database record are rolled back.
 Mongodb does not support transactions.
 
-There are currently 3 transactional callbacks that can be handled on the event, and 2 transactional callbacks for all events.
+There are currently 3 transactional callbacks that can be handled on the event, and 3 transactional callbacks for all events, and 1 transaction callback for states.
 
 ```ruby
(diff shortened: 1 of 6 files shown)


RickCSong:
@anilmaurya Could you take a look at this PR? Thank you!! :)

anilmaurya:
@RickCSong Thank you for sending PR
I will review it in couple of days.

javiercr:
Is there any chance to get this merged?

A typical use case would be having multiple events leading to the same state, and having to enqueue some job once that state is reached. Without this PR the only workaround is duplicating code in the `after_commit` hooks for the events.

anilmaurya:
Hi @javiercr, thank you for reminder

@RickCSong, thank you for this PR

Unfortunately, :after_commit AASM callback is not safe in terms of race conditions and redundant calls.
Therefore we are suggesting to use `after_commit_everywhere` gem

Changes done in this PR is not supported with current :after_commit implementation therefore I have to close this PR.
If we can implement :after_all_commits using `after_commit_everywhere` gem then I am willing to accept that PR.

Thanks.

PR #675, opened 2020-04-03
Travis: use jruby-9.1.17.0 for a 9.1

olleolleolle:
This PR configures **builds for JRuby** in a "positive" (include, not exclude) way.

- Drop the no-longer-used Travis setting `sudo: false`. See [more at the Travis blog](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).
- Update matrix to latest released JRuby 9.1, which is [9.1.17.0 April 23, 2018](https://www.jruby.org/2018/04/23/jruby-9-1-17-0.html)
- Use "include" instead of "exclude" to pick gemfiles for JRuby
- Only pick a specific JDK for JRuby
- Still 🍏 

Thanks for maintaining this library!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: ruby
 cache: bundler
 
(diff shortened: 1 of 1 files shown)


alto:
@olleolleolle Sorry, we have switched to GitHub Actions since. Thanks for your contribution (a few years ago) ❤️

PR #677, opened 2020-04-23
Add simple option for auto-generated timestamps

jaynetics:
This aims to solve #418 in a slightly simpler way than #427 - by setting up a callback that, upon entering state `foo`, checks if there is a setter `foo_at=` and if so, uses it.

I've only added tests for POROs and ActiveRecord. I suspect this will work for most ORMs, but haven't looked into that.

If this is accepted it might make sense to mention the new option in the README. I just didn't know where to put it in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
   - [Extending AASM](#extending-aasm)
   - [ActiveRecord](#activerecord)
   - [Bang events](#bang-events)
+  - [Timestamps](#timestamps)
   - [ActiveRecord enums](#activerecord-enums)
   - [Sequel](#sequel)
   - [Dynamoid](#dynamoid)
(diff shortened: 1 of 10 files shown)


Javierchik:
Let's merge this ASAP please!

anilmaurya:
Hi @jaynetics 
Sorry, I missed your PR.
I want to merge it, can you update `Readme` with `timestamps` option and also add specs in `mongoid_persistence_spec` ?

jaynetics:
@anilmaurya I've added a Readme entry and mongo spec as you requested. Let me know if anything else needs to be done!

anilmaurya:
Thank you @jaynetics for your contribution.

PR #696, opened 2020-08-11
Fix Deprecation Message About gem 'after_commit_everywhere', fix #695

### LINKED ISSUE #695: Deprecation Message About gem 'after_commit_everywhere'
**Describe the bug**
The following deprecation warning appears with the most recent gem update, regardless of whether `after_commit` is being used:
```
[DEPRECATION] :after_commit AASM callback is not safe in terms of race conditions and redundant calls.
              Please add `gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5'` to your Gemfile in order to fix that.
```

Is the `gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5'` needed _only_ if `after_commit` is being used? OR is the gem required _anytime_ the AASM gem is being used, regardless of whether `after_commit` is called or not?

**Expected behavior**
It would be helpful if the error was clearer about whether `gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5'` should be added the Gemfile only if using `after_commit` or in all cases. If `gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5'` is only needed if using `after_commit` it would be even better if the deprecation warning only appeared if using `after_commit`, and did not appear if not calling `after_commit`.

anilmaurya:


anilmaurya:
fixes #695

PR #698, opened 2020-08-11
Fix version number in the changelog

alexander-lazarov:
Noticed that the version number of the `5.1.1` release is wrong so I decided to fix it :)


anilmaurya:
Thank you 👍

PR #699, opened 2020-08-12
document how to define transitions from any state

hedgesky:
I noticed that this handy functionality isn't documented, so decided to make a quick PR fixing that.


anilmaurya:
Thank you for your contribution 👍

PR #705, opened 2020-09-03
Add Truffleruby-head to CI

gogainda:


gogainda:
I do most of my changes from Github UI so instead of rebasing I will create a new PR 
https://github.com/aasm/aasm/pull/717

PR #717, opened 2020-10-13
Add Truffleruby head to CI

gogainda:


gogainda:
Looks like truffleruby doesn't support bundler 1.16

alto:
@gogainda Sorry, we have switched to GitHub Actions since. Thanks for your contribution nonetheless! ❤️

PR #720, opened 2020-10-19
Changed migration class name to work with zeitwerk

rwegrzyniak:
Zaitwerk looks for Aasm constant because it camelize file names


anilmaurya:
Thank you @rwegrzyniak for contributing

Issue #721, opened 2020-10-22
i18n Event translations failing


fedeagripa:
**Describe the bug**
I'm trying to display transition events in an admin page, and so events translation are not working 😞 
I'm I missing something or miss reading it, or is actually not working?

**To Reproduce**
1 - Define translations for any of your event models. In my case I have `Order` with `process` event for example.
2 - Call `order.aasm.events(permitted: true).map(&:name)`

**Expected behavior**
I would expect that my call to `mymodel.aasm.events.map(&:name)` return my possible transition events translated given my `es.yml`

**Screenshots**
`Translations`
```
es:
  activerecord:
    events:
      order:
        process: 'Procesar'
        accept: 'Aceptar'
```
`Active Admin call`
```
action_item :change_status, only: :show do
    order = Order.find(params[:id])
    dropdown_menu 'Admin' do
      order.aasm.events(permitted: true).map do |state|
        item state.name, advance_admin_order_path(state: state.name)
      end
    end
  end
```
Output:
![image](https://user-images.githubusercontent.com/23194751/96905192-86798100-146e-11eb-8829-da6fd57291f9.png)
(I should be in Spanish according es.yml)

**Additional context**
I was checking `spec/en.yml` for some context about how to build my events translations.
But this is not displayed properly

anilmaurya:
@the-spectator can you help on this issue ?

the-spectator:
@fedeagripa Event localization can be accessed using `Model.aasm.huma_state_name(event_object or symbol)` .
```ruby
Order.aasm.human_event_name(:process)
```

To extend your example:
OR
```ruby
order.aasm.events(permitted: true).map do |event|
  localized_event_name = Order.aasm.human_event_name(event.name)
  item localized_event_name, advance_admin_order_path(state: state.name)
end
```

fedeagripa:
@the-spectator amazing, that worked, thanks!
Feel free to close this issue then

PR #723, opened 2020-10-26
Event localizer improvement

the-spectator:
Related to https://github.com/aasm/aasm/issues/721

## Summary

This PR adds `#to_s` method to `AASM::Core::Event` so we can directly use `Model.aasm.human_event_name(event_object)`. 
Following is the relevant code with addition to `#to_s` enables us to pass `event_object`.
https://github.com/aasm/aasm/blob/ea041603fb72b801d4944aa8b5876ac3ca5a36d2/lib/aasm/localizer.rb#L5

### DIFF
--- a/lib/aasm/core/event.rb
+++ b/lib/aasm/core/event.rb
@@ -110,6 +110,10 @@ module AASM::Core
       transitions.flat_map(&:failures)
     end
 
+    def to_s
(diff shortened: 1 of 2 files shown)


anilmaurya:
Thank you for contributing @the-spectator

PR #739, opened 2021-04-22
fix: timestamp will work with named machine

RolandStuder:
This PR makes sure timestamps work also with named machines. Which currently is not the case.

Using timestamps on a named machine like so:

```ruby
aasm :state, timestamps: true do 
```

Will result in an error:

```ruby
AASM::UnknownStateMachineError:
       There is no state machine with the name 'default' defined in TimestampsExample!
     # ./lib/aasm/aasm.rb:70:in `aasm'
     # ./lib/aasm/base.rb:271:in `block in setup_timestamps'
```

The fix is simple, in `ts_setter = "#{aasm(aasm_name).to_state}_at="`  it was just missing the aasm_name as an argument. It missing tried to use the `default` state machine which did not exist.

Tests are updated,to use the timestamps functionality with a named machine. Let me know, if you would prefer for the test to be more explicit about what it checks.

btw: Really love the addition of timestamps, thx to @jaynetics !

### DIFF
--- a/lib/aasm/base.rb
+++ b/lib/aasm/base.rb
@@ -268,7 +268,7 @@ module AASM
 
       after_all_transitions do
         if self.class.aasm(:"#{aasm_name}").state_machine.config.timestamps
-          ts_setter = "#{aasm.to_state}_at="
+          ts_setter = "#{aasm(aasm_name).to_state}_at="
           respond_to?(ts_setter) && send(ts_setter, ::Time.now)
         end
       end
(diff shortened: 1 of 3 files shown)


anilmaurya:
Hi @RolandStuder 

Thank you for reporting bug and sending PR to fix it.

Instead of updating current test, can you add a separate test for named state machine ?

RolandStuder:
@anilmaurya updated to use a separate test, I hope this is more or less what you had in mind. (force pushed btw). Thx for the quick response!

anilmaurya:
Thank you @RolandStuder for contributing    👍

PR #745, opened 2021-06-27
📝 Fix ActiveRecord spelling

jaredsmithse:
Found some inconsistent spellings for ActiveRecord while perusing the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1084,7 +1084,7 @@ end
 
 which then leads to `transaction(requires_new: false)`, the Rails default.
 
-Additionally, if you do not want any of your active record actions to be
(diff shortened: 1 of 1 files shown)


anilmaurya:
Thanks for your contribution @jaredsmithse

rafaelfranca:
Just FYI, now the spelling is wrong. `Active Record` is the right spelling of the framework.

rafaelfranca:
See https://guides.rubyonrails.org/api_documentation_guidelines.html#wording

> The proper names of Rails components have a space in between the words, like "Active Support". `ActiveRecord` is a Ruby module, whereas Active Record is an ORM.

PR #749, opened 2021-08-20
add ruby 3.0.2 to travis-ci runs

aka47:
ruby 3.0 is released and the new default ruby version. Lets ran specs against this version and fix all possible upcoming bugs.


aka47:
moving to github actions

PR #750, opened 2021-08-20
add github actions config

aka47:
As travis stopped be free for all - lets move to github actions. Also add ruby-3.0 and trufflerruby to test matrix

Fixes #749, #717 and  #675


aka47:
@anilmaurya this is ready to be merged.. can you review it or whom should I ask for this? 
The action run is visible here.. https://github.com/aka47/aasm/actions/runs/1150819885 

Once merged and configured it will show up under checks on every PR..

anilmaurya:
@aka47 Thank you for this PR

I can see that you have removed CI for nobrainer, as we support nobrainer so we should run specs for nobrainer as well.

aka47:
> @aka47 Thank you for this PR
> 
> I can see that you have removed CI for nobrainer, as we support nobrainer so we should run specs for nobrainer as well.

For some reason it got deactivated in main .. https://github.com/aasm/aasm/blob/master/.travis.yml

looks like this was the reason..
https://github.com/aasm/aasm/commit/100082e4d7b6ece2dabf2724f8e19269c825e83b#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

It looks better be done in a separate PR once this one is merged.. as it is not directly related to the switch.. , what you think?

anilmaurya:
Ohh, it was travis issue back then.

Make sense to fix in separate PR

PR #753, opened 2021-09-18
✨ Raise `AASM::UndefinedState` when calling `fire` or `fire!` with undefined event names 

danielvdao:
When firing events that are unknown, we should check the state machine if the events exist before doing so, otherwise we'll get a `NoMethodError`.

Note - I saw the original PR here: https://github.com/aasm/aasm/pull/639 and wanted to see if I could play around and get it to pass 🙂 Also I recently started using the gem, so I was looking for a way to learn the gem better by contributing. Let me know if this isn't correct and I'll be happy to close it 🙇🏽‍♂️

### DIFF
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -1,6 +1,5 @@
 module AASM
   class InstanceBase
-
     attr_accessor :from_state, :to_state, :current_event
 
     def initialize(instance, name=:default) # instance of the class including AASM, name of the state machine
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -92,6 +92,7 @@ module AASM
     def state_object_for_name(name)
       obj = @instance.class.aasm(@name).states.find {|s| s.name == name}
       raise AASM::UndefinedState, "State :#{name} doesn't exist" if obj.nil?
+
       obj
     end
 
@@ -115,12 +116,19 @@ module AASM
     end
 
     def fire(event_name, *args, &block)
+      event = @instance.class.aasm(@name).state_machine.events[event_name]
+      raise AASM::UndefinedState, "State :#{event_name} doesn't exist" if event.nil?
+
       @instance.send(event_name, *args, &block)
     end
 
     def fire!(event_name, *args, &block)
-      event_name = event_name.to_s.+("!").to_sym
-      @instance.send(event_name, *args, &block)
+      bang_event_name = "#{event_name}!".to_sym
+      event = @instance.class.aasm(@name).state_machine.events[event_name]
+
+      raise AASM::UndefinedState, "State :#{bang_event_name} doesn't exist" if event.nil?
+
+      @instance.send(bang_event_name, *args, &block)
     end
 
     def set_current_state_with_persistence(state)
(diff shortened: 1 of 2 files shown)


danielvdao:
@anilmaurya can you trigger CI build since you were the original PR reviewer? I'm guessing opening a PR doesn't kick off the build like I hoped. I ran all the unit tests locally and it looks like this passed 🙏🏽

danielvdao:
Nudge @anilmaurya. Do you mind taking a look at this?

anilmaurya:
@danielvdao Sorry for late response.

I have triggered CI build, Build is passing.

I will review this PR in couple of days.

danielvdao:
@anilmaurya sounds good, hopefully should be a small change. Thank you!

lib/aasm/instance_base.rb
```diff
@@ -92,6 +91,7 @@ def permitted_transitions
     def state_object_for_name(name)
       obj = @instance.class.aasm(@name).states.find {|s| s.name == name}
       raise AASM::UndefinedState, "State :#{name} doesn't exist" if obj.nil?
+
```
anilmaurya, line 94: I think this line addition is not required

lib/aasm/instance_base.rb
```diff
@@ -115,19 +115,25 @@ def may_fire_event?(name, *args)
...
       @instance.send(event_name, *args, &block)
     end
 
     def fire!(event_name, *args, &block)
-      event_name = event_name.to_s.+("!").to_sym
-      @instance.send(event_name, *args, &block)
+      bang_event_name = "#{event_name}!".to_sym
+      event = @instance.class.aasm(@name).state_machine.events[event_name]
+
+      raise AASM::UndefinedState, "State :#{bang_event_name} doesn't exist" if event.nil?
```
anilmaurya, line 128: I think we can avoid duplicate code from `fire` method

spec/unit/complex_example_spec.rb
```diff
@@ -90,4 +90,12 @@
...
     expect(auth.aasm.may_fire_event?(:unknown)).to be false
   end
+
+  it "should raise AASM::UndefinedState when firing unknown events" do
+    expect { auth.aasm.fire(:unknown) }.to raise_error(AASM::UndefinedState, "State :unknown doesn't exist")
+  end
+
+  it "should raise AASM::UndefinedState when firing unknown bang events" do
+    expect { auth.aasm.fire(:unknown!) }.to raise_error(AASM::UndefinedState, "State :unknown! doesn't exist")
+  end
```
anilmaurya, line 100: We should add specs to make sure that callbacks are not triggered when `UndefinedState` exception is raised

anilmaurya (changes requested):
Hi @danielvdao 

I added some review comments, let me know if you have any questions

lib/aasm/instance_base.rb
danielvdao: @anilmaurya sure. Are you suggesting I extract this into another method then?

spec/unit/complex_example_spec.rb
danielvdao: Sounds good - I'll see if I can understand how callbacks get triggered, it will be fun 🙂

danielvdao:
@anilmaurya I cleaned up a bit of lines and removed some debugger statements. I also added in more tests like you suggested. Let me know what you think.

PR #754, opened 2021-09-22
Prepend timestamps with namespace to avoid naming conflicts

mclement18:
### **Context:**
A class contains two state machines that have conflicting state names and they are both using timestamps. At least one is using a namespace.

### **Actual behavior:**
The timestamp of a particular state from the first state machine will be overridden by the transition to the conflicting state of the second state machine.

### **New behavior:**
The timestamps are prepended by the namespace when provided to avoid name collision.


j-m-harris:
+1 for this issue, a timestamps prefix would also be nice (but then isn't that what the namespace already is?)

PR #763, opened 2021-11-30
Remove internal method mention from 'Transaction support' in README.md

ovedanner:
The way `aasm_fire_event` is currently listed kind of makes it look like a callback than can be defined, while it's not. I'm not sure if people need to be aware of `aasm_fire_event` at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1012,12 +1012,11 @@ Since version *3.0.13* AASM supports ActiveRecord transactions. So whenever a tr
 callback or the state update fails, all changes to any database record are rolled back.
 Mongodb does not support transactions.
 
-There are currently 3 transactional callbacks that can be handled on the event, and 2 transactional callbacks for all events.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1012,12 +1012,11 @@ Since version *3.0.13* AASM supports ActiveRecord transactions. So whenever a tr
 callback or the state update fails, all changes to any database record are rolled back.
 Mongodb does not support transactions.
 
-There are currently 3 transactional callbacks that can be handled on the event, and 2 transactional callbacks for all events.
+There are currently 3 transactional callbacks that can be handled on the event, and 2 transactional callbacks for all events:
 
 ```ruby
   event           before_all_transactions
   event           before_transaction
-  event           aasm_fire_event (within transaction)
```
alto, line 1020: @ovedanner I do agree, the method `aasm_fire_event` shouldn't be mentioned here. But we still need an indicator to show that this is the time when the actual event is fire. Maybe replacing the line with

```
                  <event fired>
```

would help? What do you think?

PR #764, opened 2021-12-06
README: fix statement about whinyness of aasm

fwolfst:
At least that is what I understand from the example and it also sounds right"er" (ya, non-native english-speaker here ;) ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ job.run       # => raises AASM::InvalidTransition
 ```
 
 If you don't like exceptions and prefer a simple `true` or `false` as response, tell
-AASM not to be *whiny*:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,7 +106,7 @@ job.run       # => raises AASM::InvalidTransition
 ```
 
 If you don't like exceptions and prefer a simple `true` or `false` as response, tell
-AASM not to be *whiny*:
+AASM not to not be *whiny*:
```
CodingAnarchy, line 109: This would be a double negative, which reads the same as telling it to be whiny. It should either be `not to` (as it is already) or `to not`: either works in English.

PR #770, opened 2022-02-02
Fix for alias_attribute

i-krav:
I am having an issue on a legacy project with an existing column. Data migration is not an option. It has value with `-` in that case AASM won't work, since it is creating a constant with state. An option was to create a custom getter/setter and gsub the `-` character. But with the current version AASM doesn't work with a custom method, which doesn't have a column in DB. So it is fixing this. So in a couple of words, with this fix it is possible to use a custom method or alias_attribute with AASM without getting an error.


i-krav:
Closing in favour of https://github.com/aasm/aasm/pull/773 It is much simpler fix.

PR #773, opened 2022-02-04
Add possibility to skip constant creation for each state

i-krav:
Basically simpler fix for https://github.com/aasm/aasm/pull/770

With it you may use states that have hyphen in the name without constant creation error.

### DIFF
--- a/spec/unit/no_constants_spec.rb
+++ b/spec/unit/no_constants_spec.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'No Constants' do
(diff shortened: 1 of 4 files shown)


spec/unit/no_constants_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'No Constants' do
+  let(:simple) { NoConstant.new }
+
+  it 'defines constants for each state name' do
```
GMorris-professional, line 8:
```suggestion
  it 'does not define constants for each state name' do
```

PR #776, opened 2022-03-05
Fix multi-threading bug not firing success callbacks

### LINKED ISSUE #448: success callback not being invoked
Here's what I have. And I have absolutely no idea how to replicate it.

```ruby
class RmaOrder < Order
  aasm column: :status do
    state S_NEW, initial: true
    state S_COMPLETE
    state S_CANCELLED

    after_all_events do
      save!
    end

    event :complete do
      transitions from: S_NEW, to: S_COMPLETE, guard: :ready_for_completion?

      success do
        # this is not running whatsoever
        set_totals
      end
    end
  end
```

And I have another class that is exactly the same (well, I literally copy pasted it), and it invokes the `success` just like it should.

State transitions successfully, `ready_for_completion?` returns `true`. If I switch to use `after` callback, it works.

Any idea?

guewen:
`@valid_transitions` is shared across threads, when a thread consumes a
transition, another thread which should execute the same transition for
another model instance may ignore its success callback.

This change prevents to share the same hash of transition across different
model instances. Each instance will have its own hash of transitions.
The hash itself is not subject to thread race conditions as the object ids
are unique. Also, the transitions for an object_id are cleaned after the
transition have been fired, otherwise the hash would grow infinitely.

See https://github.com/aasm/aasm/issues/448#issuecomment-377922368 for
an example of code to reproduce the multi-thread bug.

Notes:

* I failed to install `appraisal` as asked in the contributing guidelines, I didn't spend too much time trying as it seems it requires an old version of bundler
* I would appreciate any help on writing tests if anyone have an idea how to test this multi-thread race condition. The code snippet in https://github.com/aasm/aasm/issues/448#issuecomment-377922368 now gives the expected result, but not quite sure how I could write a test for this

Fixes #448

### DIFF
--- a/lib/aasm/core/event.rb
+++ b/lib/aasm/core/event.rb
@@ -10,7 +10,7 @@ module AASM::Core
       @name = name
       @state_machine = state_machine
       @transitions = []
-      @valid_transitions = {}
+      @valid_transitions = Hash.new { |h, k| h[k] = {} }
       @guards = Array(options[:guard] || options[:guards] || options[:if])
       @unless = Array(options[:unless]) #TODO: This could use a better name
       @default_display_name = name.to_s.gsub(/_/, ' ').capitalize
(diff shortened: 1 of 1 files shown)


guewen:
Hello @anilmaurya, any chance you could give an approval to run the workflow, so I can check I didn't break any tests?
Thank you!

guewen:
Closed in favor of #778 (I needed to change of fork).

PR #781, opened 2022-04-01
Fix a few typos in README.md

TessHuelskamp:
These should be no-ops in terms of wording and content


anilmaurya:
Thank you @TessHuelskamp for your contribution.

PR #788, opened 2022-07-25
Cast string to symbol in fire/fire!, and fix deceptive error class and message

### LINKED ISSUE #787: Upon upgrade to 5.3.0, `fire!` requires Symbol and raises error with String
As of the 5.3.0 release, which includes #753 , `fire!` requires a symbol rather than a string. If an event name is passed as a String rather than a Symbol, it raises `AASM::UndefinedState`.

This is my excerpted setup from my application with steps to reproduce.

```ruby
class CampaignMember
  aasm column: :status, timestamps: true, whiny_persistence: true do
    state :invited

    event :invite do
    transitions to: :invited
    end
  end
end

c = CampaignMember.first
c.aasm.fire!(:invite) # works as expected
c.aasm.fire!("invite")
AASM::UndefinedState: State :invite! doesn't exist
```
I would either expect the code to raise `ArgumentError` if you intend to force the type as Symbol rather than String, or for `fire` to cast a String to a Symbol internally. My preference as a user of this library would be for silent casting to Symbol, but either choice would make sense to me.

I'm not sure if this was silently failing with a String before, or whether the changes in #753 included breaking changes.

If the maintainers agree this is an issue to be fixed then I am happy to do the work and open a pull request. If it turns out I have misunderstood the library and this is in fact an issue in my code then my apologies for the wasted time.

norman:
This PR resolves #787 by making `fire` and `fire!` allow a String or Symbol argument.

I've also changed the error that is raised when these methods are passed a String or Symbol that does not refer to a known event.

At present `fire!` raises an `AASM::UndefinedState` error when an event can not be found. The error message says: `"State :#{event_error} doesn't exist"` This is confusing and arguably incorrect because the item that can not be found is not a *state*, but rather an *event*. 

If you look at the current code I think the mismatch is readily apparent:

```ruby
def event_exists?(event_name, bang = false)
  event = @instance.class.aasm(@name).state_machine.events[event_name]
  event_error = bang ? "#{event_name}!" : event_name

  raise AASM::UndefinedState, "State :#{event_error} doesn't exist" if event.nil?
end
```
To make this more transparent to users of this library, I have added an `AASM::UndefinedEvent` and changed the error message string to reflect this difference.

To retain backwards compatibility and avoid having to release the next version as`5.4.0`,  I have made `AASM::UndefinedEvent` a subclass of `AASM::UndefinedState`. Applications that already count on rescuing `AASM::UndefinedState` in this context should still work without modification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## unreleased
 
+* Allow `fire` and `fire!` to accept a String or Symbol for the event name, and raise AASM::UndefinedEvent rather than AASM::UndefinedState
+  when an event can not be found, thanks to [norman](https://github.com/norman).
+
 ## 5.3.0
 
 * Add Ruby 3.1 and Rails 7 to the CI matrix [#775](https://github.com/aasm/aasm/pull/775), thanks to [petergoldstein](https://github.com/petergoldstein)
--- a/lib/aasm/errors.rb
+++ b/lib/aasm/errors.rb
@@ -17,5 +17,6 @@ module AASM
   end
 
   class UndefinedState < RuntimeError; end
+  class UndefinedEvent < UndefinedState; end
   class NoDirectAssignmentError < RuntimeError; end
 end
--- a/lib/aasm/instance_base.rb
+++ b/lib/aasm/instance_base.rb
@@ -134,7 +134,7 @@ module AASM
     private
 
     def event_exists?(event_name, bang = false)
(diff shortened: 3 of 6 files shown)


anilmaurya:
Thank you @norman for your contribution 👍

norman:
> Thank you @norman for your contribution 👍

My pleasure. Thanks for maintaining this library, we are making heavy use of it at work and it's saving us a lot of time and effort!

Issue #497, opened 2017-08-17
Inverted "not_" scopes


lichtamberg:
Hi,

It would be nice to have also inverted automatic scopes, like f.e. for a Sms-Model:

```ruby
state :delivered

=> SmsMessage.delivered
```

It would be nice if we would add the "not_" scopes as well:

```ruby
=> SmsMessage.not_delivered
```

If you like, I can open a PR? Just tell me if you would accept such a change or if you have any thoughts about this?

derikson:
Another option is to allow a boolean argument, `inverted` that defaults to false. Then you would use `SmsMessage.delivered(false)`. I prefer this when you have filters where the user may turn it on or off, because then you only have to change the argument instead of using `#send` to change the method name.

timlawrenz:
@h0jeZvgoxFepBQ2C are you still considering submitting a PR?

grozwalker:
Hello everyone! 

Do you have any news on this task? I think it would be convenient!

I'm ready to do PR

danielricecodes:
> Hello everyone!
> 
> Do you have any news on this task? I think it would be convenient!
> 
> I'm ready to do PR

@grozwalker - checkout #795 and upvote please :+1:

fauno:
I come here to propose something similar, but I was thinking on automatic scopes for events, something like `Model.may_event` to get all models that could transition to an event. Is it something for another issue? Not sure I can provide a PR in the short term...

PR #795, opened 2022-09-13
Add Automatic Negative Scopes

danielricecodes:
Opening a PR to see what the community reaction is to automatic negative scopes?  

With the following example state machine:

```ruby
  include AASM

  aasm do
    state :sleeping, initial: true
    state :running
    state :cleaning
  end
end
```

Would produce 6 scopes instead of the usual 3.  
  * `Job.sleeping`
  * `Job.not_sleeping`
  * `Job.running`
  * `Job.not_running`
  * `Job.cleaning`
  * `Job.not_cleaning`

Sometimes you want to query for everything except a certain state.


gildo:
Also https://github.com/rails/rails/pull/35381

tagliala:
Interesting, thanks.

Should close #497

Any chances to add specs?

danielricecodes:
> Interesting, thanks.
> 
> Should close #497
> 
> Any chances to add specs?

Thanks for the nudge.  This branch has been rebased onto current `master` and I added some specs for it.  If there are other places I need to add tests, please let me know where.

danielricecodes:
@tagliala - another thing I noticed in the source is that there appears to be an "if ActiveRecord is 3 or higher" in here and I'm not sure why it's still there.  The `else` contains Rails 2 scope syntax and this gem does not support Rails 2 anymore.  The `if/else` looks like it can be removed, unless I'm missing something.

```ruby
      module ClassMethods
        def aasm_create_scope(state_machine_name, scope_name)
          if ActiveRecord::VERSION::MAJOR >= 3
            conditions = { aasm(state_machine_name).attribute_name => scope_name.to_s }
            class_eval do
              scope scope_name, lambda { where(table_name => conditions) }
              scope "not_#{scope_name}", lambda { where.not(table_name => conditions) }
            end
          else # WHAT IS THE PURPOSE OF THIS?  MONGOID???
            conditions = {
              table_name => { aasm(state_machine_name).attribute_name => scope_name.to_s }
            }
            class_eval do
              named_scope scope_name, :conditions => conditions
            end
          end
        end
      end
 ```

tagliala (changes requested):
Hi, just a couple of minor improvements for uniformity

I'm not a contributor of this gem

tagliala:
> The else contains Rails 2 scope syntax and this gem does not support Rails 2 anymore

As a non contributor, I can't tell.

The changelog says that this gem is not compatible anymore with Rails 3 starting from 5.1.0, so I guess that it is dead code https://github.com/aasm/aasm/blob/master/CHANGELOG.md#510

I would not remove that in this PR

danielricecodes:
@rubyist - can a maintainer please enable CI on this PR, please?  Thank you :+1:

danielricecodes:
Closing. Not merged for years.

PR #823, opened 2023-05-15
Fixes #822 - Namespaced scopes perform incorrect queries

### LINKED ISSUE #822: Namespaced scopes perform incorrect queries
**Describe the bug**
Namespaced scopes were implemented in #735. They do not work the way I expect.

**To Reproduce**
Add the following to the spec for namespaced scopes:

```ruby
it "creates identical queries" do
  expect(MultipleNamespaced.car_unsold.all.to_sql == MultipleNamespaced.unsold.all.to_sql).to be_truthy
  expect(MultipleNamespaced.car_sold.all.to_sql == MultipleNamespaced.sold.all.to_sql).to be_truthy
end
```

The spec will fail, because the non-namespaced query is:

```
SELECT "multiple_namespaceds".* FROM "multiple_namespaceds" WHERE "multiple_namespaceds"."status" = 'unsold'
```

While the namespaced query is:

```
SELECT "multiple_namespaceds".* FROM "multiple_namespaceds" WHERE "multiple_namespaceds"."status" = 'car_unsold'
```

**Expected behavior**
The namespaced query should not be testing a different status value. I do not expect the status values stored in the database to be namespaced.

kranzky:
Namespaced scopes were prepending the namespace to the state name when performing the query. This is wrong. The fix is to pass the scope name and the state name as separate arguments to `aasm_create_scope`.

### DIFF
--- a/lib/aasm/persistence/active_record_persistence.rb
+++ b/lib/aasm/persistence/active_record_persistence.rb
@@ -40,15 +40,15 @@ module AASM
       end
 
       module ClassMethods
-        def aasm_create_scope(state_machine_name, scope_name)
+        def aasm_create_scope(state_machine_name, scope_name, state_name)
           if ActiveRecord::VERSION::MAJOR >= 3
-            conditions = { aasm(state_machine_name).attribute_name => scope_name.to_s }
+            conditions = { aasm(state_machine_name).attribute_name => state_name.to_s }
             class_eval do
               scope scope_name, lambda { where(table_name => conditions) }
             end
           else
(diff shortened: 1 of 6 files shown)


ccartano:
Running into the same issue -- interesting that the namespaced scopes does this since the scope then doesnt work at all. Would love to see this added in or if theres a natural solution for it would be helpful!

vad4msiu:
👍 
Could we merge it?

rafaelsales:
any hope to get this out?

PR #833, opened 2023-06-11
Allow passing an empty hash to `:after` transition methods

### LINKED ISSUE #829: Can't pass an empty hash as callback parameters to `transition_from`  since 5.5.0
**Describe the bug**

Can't pass an empty hash as callback parameters to `transition_from`  since 5.5.0

**To Reproduce**

Reproduce script is here.

```ruby 
# frozen_string_literal: true
# bug_report_aasm.rb 

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "aasm", "5.5.0"
  gem "rspec"
end

class Job
  include AASM

  aasm do
    state :sleeping, initial: true
    state :running, :cleaning

    event :run do
      transitions from: :sleeping, to: :running, after: :prepare_run!
    end

    event :clean do
      transitions from: :running, to: :cleaning
    end
  end

  def prepare_run!(params)
    puts params
  end
end

require "aasm/rspec"
require "rspec/autorun"

RSpec.describe Job do
  it "move to 'run' with parameters" do
    job = Job.new
    expect(job).to transition_from(:sleeping).to(:running).on_event(:run, {})
  end
end
```

```bash
$ ruby bug_report_aasm.rb   
F

Failures:

  1) Job move to 'run' with parameters
     Failure/Error:
       def prepare_run!(params)
         puts params
       end

     ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # bug_report_aasm.rb:28:in `prepare_run!'
     # bug_report_aasm.rb:39:in `block (2 levels) in <main>'

Finished in 0.00177 seconds (files took 0.05501 seconds to load)
1 example, 1 failure

Failed examples:

rspec bug_report_aasm.rb:37 # Job move to 'run' with parameters
```

**Expected behavior**

Allow to pass an empty hash as callback parameters via `transition_from` 

**Additional context**

This has happened since #808.

y-yagi:
Since Ruby 2.7, a double splat with an empty hash passes no arguments. So if an argument is an empty hash we should avoid it. 
Ref: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#other-minor-changes-empty-hash

Fix #829.

### DIFF
--- a/lib/aasm/core/invokers/literal_invoker.rb
+++ b/lib/aasm/core/invokers/literal_invoker.rb
@@ -31,7 +31,7 @@ module AASM
           return record.__send__(subject) if subject_arity.zero?
           return record.__send__(subject, *args) if subject_arity < 0
           req_args = args[0..(subject_arity - 1)]
-          return record.__send__(subject, **req_args[0]) if req_args[0].is_a?(Hash)
+          return record.__send__(subject, **req_args[0]) if req_args[0].is_a?(Hash) && !req_args[0].empty?
           record.__send__(subject, *req_args)
         end
         # rubocop:enable Metrics/AbcSize
--- a/spec/unit/rspec_matcher_spec.rb
+++ b/spec/unit/rspec_matcher_spec.rb
@@ -72,6 +72,11 @@ describe 'state machine' do
       expect(example).to_not allow_event(:fill_out_with_args).with(false)
     end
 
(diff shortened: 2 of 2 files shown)


y-yagi:
Hmm. Do I need a refactoring in this case?

PR #841, opened 2023-10-19
Support rails 6.0 in CI

xjunior:
5.2 is in CI, as 6.1 is. I don't see why not keep 6.0 there, either, to keep the code in check. Many companies are still migrating from 5.2 to 6.0, and then to 6.1, and this gives us the safety net we need.


xjunior:
Hello! I've run this locally, and it passes, can we start by approving the workflow?

xjunior:
Hello, @anilmaurya. I'm sorry to bother you on this OS library, but would you mind giving some love to this PR?

anilmaurya:
Thanks for contributing @xjunior

PR #849, opened 2024-04-29
Fix keyword parameters in ruby 3.2

matteomanzo:
The array argument will capture a Hash as the last entry if keywords were provided by the caller after all positional arguments.

Reference: https://docs.ruby-lang.org/en/3.2/syntax/methods_rdoc.html#label-Array-2FHash+Argument

See also #844

### DIFF
--- a/lib/aasm/core/invokers/literal_invoker.rb
+++ b/lib/aasm/core/invokers/literal_invoker.rb
@@ -31,8 +31,13 @@ module AASM
           return record.__send__(subject) if subject_arity.zero?
           return record.__send__(subject, *args) if subject_arity < 0
           req_args = args[0..(subject_arity - 1)]
-          return record.__send__(subject, **req_args[0]) if req_args[0].is_a?(Hash)
-          record.__send__(subject, *req_args)
+          if req_args[-1].is_a?(Hash)
+            positional_args = req_args[0..-2]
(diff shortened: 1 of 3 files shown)


lib/aasm/core/invokers/literal_invoker.rb
```diff
@@ -30,12 +30,21 @@ def exec_subject
           raise(*record_error) unless record.respond_to?(subject, true)
           return record.__send__(subject) if subject_arity.zero?
           return record.__send__(subject, *args) if subject_arity < 0
```
msxavi, line 32: The problem seems to come from line 32; `subject_arity` is -1 for a subject like below:
```
def set_follow_up_date(date: nil)
end
```

alto:
@matteomanzo Closed in favour of https://github.com/aasm/aasm/pull/857 . Thanks for looking into this ❤️

msxavi:
@brian-carroll for me, the work around was to replace state callback:
```
def set_follow_up_date(date: nil)
```
with:
```
  # TODO: make date mandatory
  # Temporary fix to allow for positional arguments: # reported issue with positional arguments: https://github.com/aasm/aasm/issues/845
  def set_follow_up_date(args = {})
    args[:date] && assign_attributes(follow_up_date: args[:date])
  end

```

PR #850, opened 2024-07-30
Add current state detail to minitest transition_from assertions

h-lame:
It's useful when reviewing a failure to know which state you did get by firing the transition, not just the one you expected, as this can help with tracing down the error in the state machine. This brings the minitest helper closer to the output from the rspec version which already includes the actual state not just the expected one.


h-lame:
Failure here from codeclimate seems overzealous - but I'm willing to try resolving it if a maintainer thinks there's merit in doing so.

PR #857, opened 2024-12-17
Ruby 3.2 compatibility ✅

thiagofm:
For the invoker classes, we started to rely on `.parameters` to know if there are keyword arguments. It's a clearer approach than relying on arity. I couldn't make it work by just relying on arity.

I've also improved test coverage a bit, those new test cases were all failing because of the Ruby 3.2 upgrade, so the new tests were added because of that.

### DIFF
--- a/lib/aasm/core/invokers/literal_invoker.rb
+++ b/lib/aasm/core/invokers/literal_invoker.rb
@@ -26,9 +26,20 @@ module AASM
         end
 
         def exec_subject
-          raise(*record_error) unless record.respond_to?(subject, true)
-          return record.__send__(subject) if subject_arity.zero?
+          ensure_method_exists
+          return simple_invoke if subject_arity.zero?
+          invoke_with_arguments
+        end
 
+        def ensure_method_exists
+          raise(*record_error) unless record.respond_to?(subject, true)
+        end
+
+        def simple_invoke
+          record.__send__(subject)
+        end
+
+        def invoke_with_arguments
           if keyword_arguments?
             instance_with_keyword_args
           elsif subject_arity < 0
@@ -44,18 +55,12 @@ module AASM
         end
 
         def instance_with_keyword_args
-          if args.last.is_a?(Hash)
-            new_args = args[0..-2]
-            keyword_args = args.last
-          else
-            new_args = args
-            keyword_args = nil
-          end
+          positional_args, keyword_args = parse_arguments
 
           if keyword_args.nil?
-            record.send(subject, *new_args)
+            record.send(subject, *positional_args)
           else
-            record.send(subject, *new_args, **keyword_args)
+            record.send(subject, *positional_args, **keyword_args)
           end
         end
 
--- a/lib/aasm/core/invokers/literal_invoker.rb
+++ b/lib/aasm/core/invokers/literal_invoker.rb
@@ -25,16 +25,36 @@ module AASM
           @arity ||= record.__send__(:method, subject.to_sym).arity
         end
 
-        # rubocop:disable Metrics/AbcSize
         def exec_subject
           raise(*record_error) unless record.respond_to?(subject, true)
-          return record.__send__(subject) if subject_arity.zero?
-          return record.__send__(subject, *args) if subject_arity < 0
-          req_args = args[0..(subject_arity - 1)]
-          return record.__send__(subject, **req_args[0]) if req_args[0].is_a?(Hash)
-          record.__send__(subject, *req_args)
+          
+          if subject_arity.zero?
+            record.__send__(subject)
+          elsif subject_arity < 0
+            invoke_with_variable_arity
+          else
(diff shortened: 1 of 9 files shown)


nyghtly-derek:
Thank you for working on this fix! Am I correct to assume that this will fix the same problems with keyword parameters as seen in https://github.com/aasm/aasm/pull/849?

Do you think this is ready for merge? Maybe we should ping a member of https://github.com/orgs/aasm/people?

nyghtly-derek:
@anilmaurya Pinging you as the most recent maintainer. Do you think we're good to merge this fix for ruby 3.2?

spec/models/event_with_keyword_arguments.rb
```diff
@@ -9,8 +9,16 @@ class EventWithKeywordArguments
...
     end
+
+    event :another_close do
+      before :_before_another_close
+      transitions from: :open, to: :closed
+    end
   end
 
   def _before_close(key:)
   end
```
msxavi, line 20: Can we please include a scenario where key is optional?
```
[1] pry(main)> def method_a(key:)
[1] pry(main)* end
=> :method_a
[2] pry(main)> def method_b(key: nil)
[2] pry(main)* end
=> :method_b
[4] pry(main)> method(:method_a).arity
=> 1
[5] pry(main)> method(:method_b).arity
=> -1
[6] pry(main)>
```

That should fall within `invoke_with_variable_arity` which from my understanding will fail.

thiagofm:
@nyghtly-derek

I believe it is ready and fixed, I've ran it in production.

It could be that there's some unhandled / corner case, but in case that there is, we can fix as it shows up. Right now the test suite covers a lot of them.

I did send @anilmaurya an email a few months ago, he was busy at the time.

spec/models/event_with_keyword_arguments.rb
thiagofm: My solution no longer relies on checking for arity, so it might still work.

I'll do this when I have time. If you want feel free to create a PR on my fork 💯

lib/aasm/core/invokers/literal_invoker.rb
```diff
@@ -25,16 +25,57 @@ def subject_arity
...
+
+        def invoke_with_fixed_arity
           req_args = args[0..(subject_arity - 1)]
-          return record.__send__(subject, **req_args[0]) if req_args[0].is_a?(Hash)
-          record.__send__(subject, *req_args)
+          if req_args[0].is_a?(Hash)
+            record.__send__(subject, **req_args[0])
+          else
+            record.__send__(subject, *req_args)
+          end
```
derek-noredink, line 77: Hey @thiagofm,

Thank you for working on this fix! I tried it out on my project and found what I think is a bug in your implementation. Here's the example case:

```
def sample_contents_present?(_client, sample)
  sample["content"].present?
end
```

And the relevant stack trace:
```
Error
wrong number of arguments (given 1, expected 2)
Stacktrace
./app/models/guided_draft/student_state.rb:328:in `sample_contents_present?'
.../lib/ruby/gems/3.2.0/bundler/gems/aasm-231f2fabedb4/lib/aasm/core/invokers/literal_invoker.rb:74:in `invoke_with_fixed_arity'
.../lib/ruby/gems/3.2.0/bundler/gems/aasm-231f2fabedb4/lib/aasm/core/invokers/literal_invoker.rb:48:in `invoke_with_arguments'
.../lib/ruby/gems/3.2.0/bundler/gems/aasm-231f2fabedb4/lib/aasm/core/invokers/literal_invoker.rb:31:in `exec_subject'
.../lib/ruby/gems/3.2.0/bundler/gems/aasm-231f2fabedb4/lib/aasm/core/invokers/literal_invoker.rb:19:in `invoke_subject'
.../lib/ruby/gems/3.2.0/bundler/gems/aasm-231f2fabedb4/lib/aasm/core/invokers/base_invoker.rb:48:in `invoke'
```

Our theory about why it's broken is essentially that we're missing a case for:
- fixed arity, with more than one argument, with the first argument being a hash

In this scenario, we do not want to only send the first argument, as you've done here:
```
if req_args[0].is_a?(Hash)
  record.__send__(subject, **req_args[0])
```

Rather, we want both arguments, that is:
```
record.__send__(subject, *req_args)
```

Does that make sense to you?

alto:
@thiagofm Thanks for your contribution. This is a nasty bug you had been looking into here. ❤️ 

@thiagofm  @msxavi @derek-noredink Is there anything missing still? Otherwise I'm happy to merge in and release a new version for this.

derek-noredink:
@alto Thank you for following up on this! We were actually able to upgrade to ruby 3.2 by using aasm 5.4. We only had to write a little work around for one of our tests.

In any case, I can't trace anything back to a problem with this pull request, so I'd say we're good to go. This is definitely an improvement!

